PR #38, opened 2016-10-30
Migrating to Castle ESC

evanxq:
Took the old throttling-up-and-down functionality out of motorControl.ino because we now can


barulicm:
This is being merged into the new_esc branch for now until the hardware is switched from the old ESC to the new ESC.

PR #45, opened 2017-01-24
Path Planning

evanxq:
Now with predictive path integration


avc/src/planner/planner.cpp
```diff
@@ -1,129 +1,199 @@
...
+    pathMsg.header.frame_id = "map";
+    path_pub.publish(pathMsg);
+}
+
+//a convenient scope
+void planner::spin() {
+    normal_distribution<double> nd(0, STEER_STDDEV);
+    steering_gaussian_ptr = &nd;
+    default_random_engine dre;
+    rand_gen_ptr = &dre;
```
barulicm: *warning:* You have entered the realm of random number generation, so I will have a lot of comments here.

1. Why do these need to be pointers?
2. Why can't they just be initialized in the constructor after you've read the *STEER_STDDEV* parameter?
3. std::default_random_engine is implementation defined, which is sketchy as it could change planner performance based on the compiler we use. *mt19937* and *minstd_rand0* are both popular generators.
4. Standard generators take a seed value to their constructor. By not providing a seed value, you are forcing the generator to use its default seed. This actually means you're using the same seed every time the code is run, and you're numbers aren't random at all, but just a very convoluted way to get the same sequence of numbers every time you run. A better (not perfect, but good enough) solution is to pass `std::random_device{}()` to the generators constructor. We can talk about all the details going on here at the next meeting.

avc/src/planner/planner.cpp
```diff
@@ -1,129 +1,199 @@
...
+
+    vector<double> bestSteering;
+    for(int i = 0; i < weightedSteeringRaw.size(); i++) {
+        bestSteering.push_back(weightedSteeringRaw[i] / weightTotal);
+    }
+
+    sim_path bestPath = calculatePath(bestSteering);
+
+    desired_steer_angle = bestPath.angles[0];
+    desired_velocity = steeringToSpeed(desired_steer_angle);
```
barulicm: A small optimization, but isn't this speed already calculated at `bestPath.speeds[0]` ?

avc/src/planner/planner.cpp
```diff
@@ -1,129 +1,199 @@
...
+    speed_pub.publish(speedMSG);
+    steer_pub.publish(steerMSG);
+
+    nav_msgs::Path pathMsg;
+    for(int i = 0; i < bestPath.poses.size(); i++) {
+        geometry_msgs::PoseStamped p;
+        p.pose.position.x = bestPath.poses[i].x;
+        p.pose.position.y = bestPath.poses[i].y;
+        pathMsg.poses.push_back(p);
+    }
```
barulicm: Consider using `std::transform` for this algorithm.

avc/src/planner/planner.cpp
```diff
@@ -1,129 +1,199 @@
...
+    int nPathPoses;
+    weightTotal = nPathPoses = 0;
+    vector<double> weightedSteeringRaw;
+    for(int i = 0; i < PATH_ITERATIONS; i++) {
+        vector<double> steerPath;
+        for(int s = 0; s < PATH_STAGES; s++) {
+            angle = steeringSample();
+            for(double t = 0; t <= PATH_STAGE_TIME; t += TIME_INCREMENT) {
+                steerPath.push_back(angle);
+            }
```
barulicm: Consider using `std::fill_n` for this algorithm.

avc/src/planner/planner.h
```diff
@@ -53,9 +52,20 @@ class planner {
...
-	double calculatePathCost(double velocity, double steer_angle, pcl::PointCloud<pcl::PointXYZ>::Ptr Map);
-	int costAtPose(pose step, pcl::PointCloud<pcl::PointXYZ>::Ptr Map);
+	struct sim_path
+	{
+		std::vector<double> angles;
+		std::vector<pose> poses;
+		std::vector<double> speeds;
+	};
+
+	pose calculateStep(double speed, double steer_angle, double timestep);
```
barulicm: Instead of making an overload for this function, just give pStart a default value in the full signature.
ie: `pose pStart = pose{0,0,0}`

avc/src/mapper/mapper.cpp
```diff
@@ -68,7 +68,7 @@ int main(int argc, char** argv) {
     pcl::VoxelGrid<pcl::PointXYZ> filter;
     filter.setLeafSize(0.2f, 0.2f, 0.2f);
 
-    ros::Rate rate(20);
+    ros::Rate rate(60);
```
evanxq: Mapper is publishing at 15hz on my machine whether the rate is set to 20 or 60, so I changed it back. I must have mistakenly suspected at some point that it could publish faster.

avc/launch/avc.launch
```diff
@@ -1,9 +1,7 @@
 <launch>
-    <node pkg="avc" type="color_detector_avc" name="color_detector" output="screen"/>
-    <!-- <param name="/use_sim_time" value="true"/> -->
-
     <include file="$(find rr_platform)/launch/transform.launch"/>
     <include file="$(find avc)/launch/mapper.launch"/>
     <include file="$(find avc)/launch/planner.launch"/>
     <include file="$(find rr_description)/speedi/speedi_state_publisher.launch"/>
+    <node pkg="rr_platform" type="scanToPointCloud" name="scanToPointCloud" output="screen" />
```
barulicm: Thanks for moving scanToPointcloud to rr_platform, but this line should still live back in lidar.launch. This node can be run manually when running on bag data. If we start accumulating a lot of nodes that need to be run manually for bag files, we'll make a new launch file for that.

avc/src/planner/planner.cpp
```diff
@@ -1,129 +1,188 @@
...
+
+    map_sub = nh.subscribe("map", 1, &planner::mapCb, this);
+    speed_pub = nh.advertise<rr_platform::speed>("speed", 1);
+    steer_pub = nh.advertise<rr_platform::steering>("steering", 1);
+    path_pub = nh.advertise<nav_msgs::Path>("path", 1);
+
+    normal_distribution<double> nd(0, STEER_STDDEV);
+    steering_gaussian = nd;
+    mt19937 rand_engine(std::random_device{}());
+    rand_gen = rand_engine;
```
barulicm: These initializations don't need to be broken into multiple lines.
You can just write `steering_gaussian = normal_distribution<double>(0, STEER_STDDEV);`

PR #49, opened 2017-01-25
reconciles tf tree changes for avc

evanxq:
Can now run avc.launch from bag file. The dual joint state updater issue as a separate pull request once it's done.


barulicm:
Merged

PR #63, opened 2017-02-12
"suppresses" "non-maximal" clusters of paths

evanxq:


avc/src/planner/planner.cpp
```diff
@@ -126,39 +177,107 @@ void planner::mapCb(const sensor_msgs::PointCloud2ConstPtr& map) {
...
+
+        if(wsv.weight > bestWeight) bestWeight = wsv.weight;
+    }
+    //ROS_INFO("best weight %f", bestWeight);
+
+    /* Filter paths by weight and sort them by first steering value.
+     * Store the results in weightedSteerVecsFiltered.
+     */
+    vector<WeightedSteeringVec> weightedSteerVecsFiltered;
+    for(int i = 0; i < weightedSteerVecs.size(); i++) {
```
barulicm: If you're not using the index in your calculations, consider using a [range-based for loop](http://en.cppreference.com/w/cpp/language/range-for).

avc/src/planner/planner.cpp
```diff
@@ -126,39 +177,107 @@ void planner::mapCb(const sensor_msgs::PointCloud2ConstPtr& map) {
...
+    //sort(weightedSteerVecsFiltered.begin(), weightedSteerVecsFiltered.end(), steeringVecCompare);
+
+    /* connect components based on distance. I refer to any random
+     * path sample as a steerVec here. The variable groups holds the
+     * sets of connected (connected = "close enough") paths
+     */
+    vector<SteeringGroup> groups;
+    // outer loop: iterate through good enough paths (around 10% of paths,
+    // depending on circumstances)
+    for(int i = 0; i < weightedSteerVecsFiltered.size(); i++) {
```
barulicm: Same recommendation as line 209

avc/src/planner/planner.cpp
```diff
@@ -126,39 +177,107 @@ void planner::mapCb(const sensor_msgs::PointCloud2ConstPtr& map) {
...
+        const WeightedSteeringVec &thisSteerVec = weightedSteerVecsFiltered[i];
+        bool foundMatch = false;
+        int thisGroupIndex;
```
barulicm: Try to avoid this* and that* variable names. Prefer names that better describe the functional or semantic difference between the two variables.

avc/src/planner/planner.h
```diff
@@ -54,12 +56,31 @@ class planner {
 
 	struct sim_path
 	{
-		std::vector<double> angles;
 		std::vector<pose> poses;
 		std::vector<double> speeds;
 	};
 
-	static geometry_msgs::PoseStamped plannerPoseToPoseStamped(pose p);
+	struct SteeringGroup;
```
barulicm: No need to forward declare this struct.

PR #68, opened 2017-04-29
Improves performance and organization of path clustering algo

ebretl:
side notes: 1) deletes old "steerer" nodes for iarrc 2) adds steering values visualizer


rr_common/src/planner/flann_test.cpp
```diff
@@ -0,0 +1,43 @@
+#include "flann/flann.hpp"
```
barulicm: Is this test application something we need in the repo?

barulicm:
Merged as part of #81

PR #77, opened 2017-06-24
New esc sahit

chsahit:
Added some optimizations to camera - it's a nodelet now and we're using image_transport


barulicm:
Merged without review / passing tests because our git structure has become a bit muddled and CircleCI is failing for un-related reasons.

PR #82, opened 2017-07-02
fixes some bugs for iarrc color detection

ebretl:
Most importantly, the cv::cvtColor call is actually made now.
Drag race launch file had not been changed to use a nodelet. 
Also changes some color detection thresholds, including implementing white. Detecting the center line of the drag race still doesn't work well because gazebo adds a weird distortion/focus/blur effect to the camera


iarrc/src/color_detector/color_detector.cpp
```diff
@@ -40,12 +38,11 @@ namespace iarrc {
...
 
-        const Mat &frame = cv_ptr->image;
-
-        Mat output = Mat::zeros(frame.rows, frame.cols, CV_8UC3);
-
-        const Mat &frame_masked = frame(mask);
-
+        const Mat &frameBGR = cv_ptr->image;
+        const Mat frameHSV = Mat::zeros(frameBGR.rows, frameBGR.cols, CV_8UC3);
+        cvtColor(frameBGR, frameHSV, CV_BGR2HSV);
```
barulicm: Minor comment for possible future work. Depending on the size of the mask, reversing the order of these operations may give better performance. Calling cvtColor on the masked frame will have to convert fewer pixels between color spaces.

PR #83, opened 2017-07-04
Firmware updates

chsahit:
All the ROS side changes to firmware - how IO is done with the arduino and through the rest of the ROS graph and how the messages are parsed.


rr_platform/msg/chassis_state.msg
```diff
@@ -0,0 +1,5 @@
+# Message for defining the state of the car
+Header header           # header for when the message is published
+bool estop              # whether the estop is on or not
```
barulicm: With boolean flags like this, it is very helpful to specify in the comments the specific meaning of each state (true/false).
Generic descriptions like this can be confusing for other developers. For example, here, does the estop being "on" mean the car is enabled or disabled? It is easy to argue for either interpretation given just this description.

rr_platform/msg/chassis_state.msg
```diff
@@ -0,0 +1,5 @@
...
+bool mux_automatic      # whether the mux is set to autonomous or human controlled
+float32 speed           # speed of the car
```
barulicm: Specify units

rr_platform/src/motor_relay_node/motor_relay_node.cpp
```diff
@@ -1,15 +1,22 @@
...
-int desiredSpeed = 0;
-int desiredSteer = 0;
-int prevAngle = 0;
-int prevSpeed = 0;
+double desiredSpeed = 0;
+double desiredSteer = 0;
+double prevAngle = 0;
+double prevSpeed = 0;
+
+double p_const = 0;
```
barulicm: Using const in the name here may be a bit misleading. I understand what you're going for, but const has specific meaning in C++. A very common naming convention for these coefficients is Kp, Ki, and Kd.

rr_platform/src/motor_relay_node/motor_relay_node.cpp
```diff
@@ -47,14 +54,32 @@ void SteeringCallback(const rr_platform::steering::ConstPtr &msg) {
 }
 
 void sendCommand(boost::asio::serial_port &port) {
-    std::string message = "$" + std::to_string(static_cast<char>(desiredSpeed)) + ", " +
-                          std::to_string(static_cast<char>(desiredSteer)) + "\n";
+    std::string message = "$" + std::to_string(desiredSpeed) + ", " +
+                          std::to_string(desiredSteer) + "," +
+                          std::to_string(p_const) + "," +
+                          std::to_string(i_const) + "," +
+                          std::to_string(d_const) + "," + "\n";
```
barulicm: Is there a reason for the extra comma at the end of the message?

rr_platform/src/motor_relay_node/motor_relay_node.cpp
```diff
@@ -47,14 +54,32 @@ void SteeringCallback(const rr_platform::steering::ConstPtr &msg) {
...
     
     try {
         boost::asio::write(port, boost::asio::buffer(message.c_str(), message.size()));
     } catch (boost::system::system_error &err) {
         ROS_ERROR("%s", err.what());
     }
+    ROS_INFO_STREAM("sent: " + message);
```
barulicm: We probably don't need to keep this kind of debugging printout around. If there are messages you frequently look back on to debug specific problems with the firmware, consider ROS_DEBUG_STREAM as an alternative which will expose the output to ROS (for reading via rqt_console or rosbag), but will, by default, hide the output from the terminal.
For more information on ROS logging levels and how to turn on debug output, see these pages:
http://wiki.ros.org/Verbosity%20Levels
http://wiki.ros.org/rosconsole#Configuration

rr_platform/src/motor_relay_node/motor_relay_node.cpp
```diff
@@ -96,6 +121,12 @@ int main(int argc, char **argv) {
     nhp.param(std::string("steering_topic"), steering_topic_name, std::string("/steering"));
     ros::Subscriber steering_sub = nh.subscribe(steering_topic_name, 1, SteeringCallback);
 
+    nhp.param(std::string("p_const"), p_const, 0.0);
```
barulicm: While these default values are sensible for the type of the variable, they are pretty useless in practice. 
Your ROS parameters should have useful default values. Under normal conditions, average users should be able to get at least minimal performance out of your node with default parameter values. Here, that might mean poorly tuned, but still functional PID constants (like P = 1 or something similar).
If there truly exist no useful default values for a parameter, then it should be a required parameter with no default value specified.

rr_platform/src/motor_relay_node/motor_relay_node.cpp
```diff
@@ -104,7 +135,7 @@ int main(int argc, char **argv) {
     nhp.param(std::string("serial_port"), serial_port_name, std::string("/dev/ttyACM0"));
     boost::asio::io_service io_service;
     boost::asio::serial_port serial(io_service, serial_port_name);
-    serial.set_option(boost::asio::serial_port_base::baud_rate(115200));
+    serial.set_option(boost::asio::serial_port_base::baud_rate(9600));
```
barulicm: Is there a reason for the drop in baud rate here?

rr_platform/src/motor_relay_node/motor_relay_node.cpp
chsahit: kP or Kp?
I've only see the first before.

rr_platform/src/motor_relay_node/motor_relay_node.cpp
chsahit: It was changed to match the arduino, where 9600 baud seemed pretty common. but we can push the baud in the firmware higher (https://arduino.stackexchange.com/questions/296/how-high-of-a-baud-rate-can-i-go-without-errors) so I am open to doing so.

rr_platform/msg/chassis_state.msg
```diff
@@ -1,5 +1,5 @@
 # Message for defining the state of the car
 Header header           # header for when the message is published
-bool estop              # whether the estop is on or not
+bool estop_on           # whether the estop is turned on
```
barulicm: Again, what does it mean for the estop to be "on"? The description here should specify how true/false maps to the robot being enabled/disabled. (same thing for the mux_automatic variable btw.

chsahit: I'll change the comment, I don't know how, if at all, i should change the variable name to be more semantically meaningful

PR #89, opened 2017-08-07
Planner rewrite with performance and legibility improvments

ebretl:
- Uses non-minimal (cost) suppression instead of density clustering and averaging
- Less code, easier to understand
- Selection of path options requires fewer control samples, meaning fewer path cost calculations and much faster execution time


rr_common/src/planner/planner.h
```diff
@@ -1,73 +1,56 @@
...
-	path::pose calculateStep(double speed, double steer_angle, double timestep, path::pose pStart = path::pose{0,0,0});
-	double steeringToSpeed(double angle);
-	double steeringSample();
-	path::path calculatePath(std::vector<float> angles);
-	double calculatePathCost(path::path path, pcl::KdTreeFLANN<pcl::PointXYZ> kdtree);
-	double costAtPose(path::pose step, pcl::KdTreeFLANN<pcl::PointXYZ> kdtree);
-	void mapCb(const sensor_msgs::PointCloud2ConstPtr& map);
-
+#include "flann/flann.hpp"
+//#include <chrono>
```
barulicm: Dead include

ebretl:
updated

ebretl:
Updated again with algorithm and parameters changed to make the bot drive straight across the intersection of the gazebo AVC track

PR #93, opened 2017-09-14
Gazebo and mapping improvements

ebretl:
* moves avc mapper to rr_common and adds more filtering power and tf integration
* adds a multi-frame map tracker in avc (does not give impressive results, this is disabled right now)
* fixes gazebo IMU
* adds gazebo depth camera
* moves lidar slightly in urdf so that the robot cuts off fewer rays in gazebo


barulicm:
I'm going to merge this in pending the latest runs of CI. 
Before AVC, we really need to get the Bigoli and Macaroni config files separated and setup correctly.

Issue #94, opened 2017-09-18
AVC start signal detector


chsahit:
Implement a detector for the start of the race in AVC, a shift from a red sign to a green sign.

PR #96, opened 2017-09-28
hough circles code compiles

chsahit:
Closes issue #94


avc/src/start_detector/start_detector.cpp
```diff
@@ -0,0 +1,67 @@
...
+            return;
+        }
+        
+        const Mat &frameBGR = cv_ptr->image;
+        Mat blurredImage;
+        GaussianBlur(frameBGR, blurredImage, Size{7, 7}, 3);
+        Mat redMask = Mat::zeros(blurredImage.rows, blurredImage.cols, CV_8U);
+        inRange(blurredImage, red_low, red_high, redMask);
+        Mat redImage;
+        bitwise_and(blurredImage, blurredImage, redImage, redMask);
```
barulicm: This method definitely ANDs the image with itself and then applies the mask.
The desired behavior can be more cleanly achieved with this:
```
blurredImage.copyTo(redImage, redMask);
```

avc/src/start_detector/start_detector.cpp
```diff
@@ -0,0 +1,67 @@
...
+        vector<Vec3f> circles;
+        HoughCircles(gray, circles, CV_HOUGH_GRADIENT, 1, 150, 10, 25, 50, 150);
+        if (circles.size() != 0) {
+            int maxSum = 0, maxX = 0, maxY = 0, maxR = 0;
+            for (size_t i = 0; i < circles.size(); i++) {
+                Point center(cvRound(circles[i][0]), cvRound(circles[i][1]));
+                Mat circleMask = Mat::zeros(gray.rows, gray.cols, CV_8U);
+                circle(circleMask, center, circles[i][2], Scalar(255, 255, 255), -1); 
+                Mat circleImage;
+                bitwise_and(redImage, redImage, circleImage, circleMask);
```
barulicm: Same comment as above about masking, but there is also probably just a better way to do this specific sum over the masked portions. Unfortunately, the sum() function doesn't take a mask.

avc/src/start_detector/start_detector.cpp
```diff
@@ -0,0 +1,67 @@
...
+                int imgSum = sum(circleImage)[0];
+                if (imgSum > maxSum) {
+                    maxSum = imgSum;
+                }
+            }
+
+            if (maxSum > 100) {
```
barulicm: We may want to actually detect the transition from red to green. 
Thinking about things like "what if someone walks in front of the sign?" and similar situations.
This might just be a thing we note for future work.

avc/src/start_detector/start_detector.cpp
```diff
@@ -0,0 +1,67 @@
...
+                start.data = true;
+                start_pub.publish(start); 
+            }
+            ROS_INFO_STREAM("Max Sum: " << maxSum << "\n");
+        }
+    }
+
+    void start_detector::onInit() {
+        NodeHandle nh = getNodeHandle();
+        NodeHandle pnh = getPrivateNodeHandle();
```
barulicm: Is `pnh` ever used?

avc/src/start_detector/start_detector.h
```diff
@@ -0,0 +1,22 @@
+#include <ros/ros.h>
+#include <sensor_msgs/Image.h>
+#include <nodelet/nodelet.h>
+#include <image_transport/image_transport.h>
+#include <cv_bridge/cv_bridge.h>
+#include <std_msgs/Bool.h>
```
barulicm: This isn't a huge deal, but the last two includes here are not needed in this file. They should be moved to the implementation file.

avc/src/start_detector/start_detector.cpp
```diff
@@ -0,0 +1,67 @@
...
+        if (circles.size() != 0) {
+            int maxSum = 0, maxX = 0, maxY = 0, maxR = 0;
+            for (size_t i = 0; i < circles.size(); i++) {
+                Point center(cvRound(circles[i][0]), cvRound(circles[i][1]));
+                Mat circleMask = Mat::zeros(gray.rows, gray.cols, CV_8U);
+                circle(circleMask, center, circles[i][2], Scalar(255, 255, 255), -1); 
+                Mat circleImage;
+                bitwise_and(redImage, redImage, circleImage, circleMask);
+                int imgSum = sum(circleImage)[0];
+                if (imgSum > maxSum) {
```
barulicm: If you're not actually keeping track of which circle gave the max output, this can be replaced by `maxSum = max(maxSum, imgSum);`

avc/src/start_detector/start_detector.cpp
```diff
@@ -0,0 +1,67 @@
+#include "start_detector.h"
+#include <pluginlib/class_list_macros.h>
+
+using namespace std;
+using namespace ros;
+using namespace cv;
+
+namespace avc {
+    const Scalar red_low{0, 0, 100};
```
barulicm: It'd be cool to name and parameterize these magic number constants (the color thresholds, Hough parameters, and sum threshold) so they can be easily tuned at competition if needed.

barulicm:
@chsahit Didn't you have to add run/build depends to a package manifest to get this to launch correctly?

avc/package.xml
```diff
@@ -44,5 +46,10 @@
   <run_depend>image_transport</run_depend>
   <run_depend>message_generation</run_depend>
   <run_depend>message_runtime</run_depend>
+  <run_depend>image_transport</run_depend>
```
chsahit: @barulicm these were the run and build depends that prevented me from being able to launch the nodelets.

barulicm: :+1: No idea why these didn't show up during my last review.

barulicm:
Ready to merge pending CI completion after updating from master.

PR #106, opened 2017-10-31
updated licenses and maintainer (I'm so productive)

chsahit:
Someone review this even though not strictly necessary


barulicm:
We should add an MIT LICENSE file to the repo as well.

chsahit:
@barulicm did the thing.

Issue #112, opened 2017-11-09
Setup catkin for rviz plugins


JasonGibson274:
Setup and example for freshmen to start working on their plugins as freshmen projects.

ebretl:
Plugin ideas: 
- speedometer
- steering wheel visualizer
-

PR #113, opened 2017-11-12
RVIZ Plugins Package

barulicm:
Adds package for rviz plugins with example panel for new members to build from.
Fixes issue #112 
@JasonGibson274 Take a look through this one. If we have any issues with it, I'll fix it here and then duplicate the setup for IGVC.


JasonGibson274:
Should I be able to look at the example plugin?

barulicm:
If you have the RR repo in your catkin workspace (and have built it), you should be able to run rviz and open the example panel by doing this:
Panels->Add New Panel->rr_rviz_plugins->ExamplePanel, OK
The example panel will be docked somewhere usually on the left or right side of the window and simply displays the speed published on /speed in m/s.

JasonGibson274:
Works for me. Is there a recommended qt guide that you recommend?

barulicm:
The official documentation (http://doc.qt.io/) is really good. I haven't really looked for any tutorials or anything outside of these docs.

jgkamat:
If you want a hands on qt demo, you could use the [robocup one we use every year](https://github.com/RoboJackets/qt-example).

barulicm:
The RoboCup one goes into detail on parts of Qt that we don't use. RR / IGVC members will just need to know how to make widgets as rviz and catkin take care of the application management and build process.

JasonGibson274:
I think we should also add a small guide so people know how to add the plugin to rviz and the basics of what they should be generating

ebretl:
Nice, works on my computer as well. Putting the header file in the includes folder is different than the rest of our repo -- any reason why you did it that way?

barulicm:
@ebretl We're actually building a class library in this package, which rviz loads at runtime. Having the headers broken out into their own directory is the correct convention when building a library.

barulicm:
@JasonGibson274 See the summary doc and comments I added. Do you think this is enough info to get them started?

rr_rviz_plugins/src/ExamplePanel.cpp
```diff
@@ -0,0 +1,48 @@
...
+/*
+ * Just as in the header, everything needs to happen in the rr_rviz_plugins namespace.
+ */
+namespace rr_rviz_plugins {
+
+ExamplePanel::ExamplePanel(QWidget *parent)
+  : rviz::Panel(parent) // Base class constructor
+{
+  // Panels are allowed to interact with NodeHandles directly just like ROS nodes.
+  ros::NodeHandle handle;
```
ebretl: This section uses 2-space indents while the rest of the PR (and code base) uses 4 spaces

barulicm: Fixed

PR #114, opened 2017-11-14
updated the example panel into a panel that publishes pose

chsahit:


chsahit:
Closing PR because we'll have real panels soon.

Issue #64, opened 2017-03-02
Setup Style Enforcement


barulicm:
Let's setup some automatic formatting and style enforcement based on [ROS's C++ style guides](http://wiki.ros.org/CppStyleGuide). There is a clang-format file available [here](https://github.com/davetcoleman/roscpp_code_format).

This should also include an [EditorConfig](http://editorconfig.org/) file matching the ROS guidelines.

barulicm:
These style guides should include the target naming convention established in #90

PR #123, opened 2018-01-15
added .editorconfig

JasonGibson274:
I highly recommend editorconfig since it supports most common editors. This is a first pass so please let me know if you want anything else specific especially on the python config since RR uses it more than IGVC.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,4 +1,4 @@
-# IGVC base editorconfig
+# RoboRacing base editorconfig
 
 # Editorconfig files are recursive, so if you want to add specifc rules for
 # your subdirectory, make a file like this in that directory.
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,55 @@
+# IGVC base editorconfig
+
+# Editorconfig files are recursive, so if you want to add specifc rules for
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -0,0 +1,55 @@
...
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+indent_style = space
+indent_size = 4
+
+# Use 4 spaces for most files
+indent_style = space
+indent_size = 2
```
jgkamat: You have two `indent_size` directives in this block

.editorconfig
```diff
@@ -0,0 +1,55 @@
...
+
+# Use 2 spaces for travis files
+[{package.json,*.yml}]
+indent_style = space
+indent_size = 2
+
+# Makefiles are required to use tabs
+[{makefile, Makefile}]
+indent_style = tab
+indent_size = 4
```
jgkamat: I don't think `indent_size` is needed here, since it dosen't apply with tabs. People can override the tab width in their own editorconfigs :smile:.

barulicm:
Related Issue: #64

Issue #115, opened 2017-11-19
Add reset command to navigtation_controller


barulicm:
Sometimes, when setting up the robot, things may happen in front of the camera that the robot mistakes for the start signal. We should add a reset command to navigation_controller so we can go back to waiting for the start signal without restarting the software.

**Details**
1. Add subscriber to navigation_controller that subscribes to `/navigation_controller/reset`. The message type can simply be `std_msgs/Empty`.
2. In the callback for this new subscriber, set the current state back to `WAITING_FOR_START`.

Issue #116, opened 2017-11-19
Rider Control Panel (rviz)


barulicm:
Create an rviz panel plugin that will serve as a place for us to put buttons for the rider to control general parts of the autonomous system. As a start, create a button that sends a message on the `/navigation_controller/reset` topic to restart the navigation controller's state machine.

**Details**

1. Create a new rviz Panel. See the example plugin in `rr_rviz_plugins` package
2. In the rviz panel, create a button that fills the width of the panel and is labelled "Reset Race State".
3. Create a Qt slot for the button that publishes a `std_msgs/Empty` to the `/navigation_controller/reset` topic.

PR #125, opened 2018-02-04
Add reset signal

Holben888:
The /resetSignal param was added to handle requests to reset the navigation state to WAITING_FOR_START. A button was added as an Rviz panel for this function. #115 #116

### DIFF
--- a/rr_common/src/navigation_controller/navigation_controller.cpp
+++ b/rr_common/src/navigation_controller/navigation_controller.cpp
@@ -1,9 +1,9 @@
 #include <ros/ros.h>
 #include <rr_platform/speed.h>
 #include <rr_platform/steering.h>
+#include <rr_platform/race_reset.h>
 #include <std_msgs/Bool.h>
 #include <std_msgs/Int8.h>
-#include <std_msgs/Empty.h>
 
 const int WAITING_FOR_START = 0;
 const int RUNNING_PLANNER = 1;
@@ -73,7 +73,7 @@ void finishLineCB(const std_msgs::Int8::ConstPtr &int_msg) {
     finishLineCrosses = int_msg->data;
 }
 
-void resetCB(const std_msgs::Empty &empty_msg) {
+void resetCB(const rr_platform::race_reset &reset_msg) {
     state = WAITING_FOR_START;
     raceStarted = false;
     updateState();
--- a/rr_common/src/navigation_controller/navigation_controller.cpp
+++ b/rr_common/src/navigation_controller/navigation_controller.cpp
@@ -119,6 +119,7 @@ int main(int argc, char** argv) {
         steerMsg.angle = steering;
         steerMsg.header.stamp = ros::Time::now();
         steerPub.publish(steerMsg);
+        ROS_INFO("Current state: %d", state);
 
         rate.sleep();
     }
--- a/rr_common/src/navigation_controller/navigation_controller.cpp
+++ b/rr_common/src/navigation_controller/navigation_controller.cpp
@@ -70,6 +70,11 @@ void finishLineCB(const std_msgs::Int8::ConstPtr &int_msg) {
     finishLineCrosses = int_msg->data;
 }
 
+vold resetCB(const std_msgs::Empty) {
+    state = WAITING_FOR_START;
+    updateState();
+}
+
 int main(int argc, char** argv) {
     ros::init(argc, argv, "navigation_controller");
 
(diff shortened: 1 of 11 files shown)


rr_common/src/navigation_controller/navigation_controller.cpp
```diff
@@ -3,12 +3,15 @@
 #include <rr_platform/steering.h>
 #include <std_msgs/Bool.h>
 #include <std_msgs/Int8.h>
+#include <std_msgs/Empty.h>
```
barulicm: Try not to directly use primitive message types (see the [std_msgs documentation](http://wiki.ros.org/std_msgs#Overview) ). I know we break this rule a low (including in this file), but since we're adding new code with this PR, let's try to get the code headed in the right direction.
In this specific case, you can add a new (blank) *.msg file to rr_platform/msg. Give it a meaningful name (ie. RaceReset), and don't forget to add it to the CMakeLists file for that package.

The reasoning behind this is, as stated in the referenced docs above, primitive message types convey no semantic meaning (especially the empty type). In addition, using primitive message types weakens the type checking that ROS does for us when pairing up publishers and subscribers. ROS will happily connect any pub/sub pair using the Empty message type if we make a mistake. Making our own message types will only allow ROS to pair up publishers and subscribers that want to work with the right data.

rr_rviz_plugins/CMakeLists.txt
```diff
@@ -31,11 +31,11 @@ include_directories(
 )
 
 set (rviz_plugins_SRCS
-    src/ExamplePanel.cpp
+    src/ResetPanel.cpp
```
barulicm: Please don't replace the ExamplePanel, just add your new panel along side it.

rr_rviz_plugins/plugin_description.xml
```diff
@@ -1,9 +1,9 @@
 <library path="lib/librr_rviz_plugins">
-    <class name="rr_rviz_plugins/ExamplePanel"
-           type="rr_rviz_plugins::ExamplePanel"
+    <class name="rr_rviz_plugins/ResetPanel"
```
barulicm: Same issue as in CMakeLists.txt, don't replace the ExamplePanel

rr_rviz_plugins/include/rr_rviz_plugins/ResetPanel.h
```diff
@@ -0,0 +1,30 @@
...
+ */
+namespace rr_rviz_plugins {
+
+class ResetPanel : public rviz::Panel {
+Q_OBJECT
+public:
+    ResetPanel(QWidget *parent = 0);
+
+protected:
+    ros::NodeHandle nh;
```
barulicm: Consider matching the ExamplePanel a little more closely. You don't actually need to save _nh_ or _reset_btn_ beyond the scope of the constructor, so they don't need to be member variables of the class. Also, while default initializers (as used here for _reset_pub_ are awesome, it's generally cleaner to only use the for trivially construct-able types.

barulicm:
Also, with regards to creating PRs. If your PR addresses one or more existing issues in the repo, please tag them in the description.
In this case, this PR resolves #115 and #116

rr_common/src/navigation_controller/navigation_controller.cpp
barulicm: I realize that the instructions I added to the GitHub issue you're working on specified to use std_msgs/Empty. This was a mistake on my part.

Holben888:
Ah, thanks for the thorough feedback @barulicm. I'm hacking through this based on my (very) limited knowledge of C++, so I'm glad to pick up some best practices.

rr_rviz_plugins/include/rr_rviz_plugins/ResetPanel.h
Holben888: Strangely, moving any of the three statements (declaring the button, declaring the NodeHandle, and advertising) out of the header and into the .cpp causes the button to stop working. I've tried all combinations and the only thing that works is keeping it as is. This site enforces the fact for creating QPushButtons: https://wiki.qt.io/How_to_Use_QPushButton

rr_platform/msg/race_reset.msg
```diff
@@ -0,0 +1 @@
+# Message for reseting the state in the navigation controller
```
JasonGibson274: @barulicm @chsahit @ebretl do we want to start requiring time stamps on messages as a convention?

barulicm: I don't think so. I saw that pop up on your repo, but honestly there are a lot of message types that we don't care about timestamping.
We're not going to be syncing this data up with any other sources, so the timestamp would never be used.

PR #134, opened 2018-02-18
Ultrasonic Array

btdubs:
Adds ultrasonic array interface node for new ultrasonic hardware to take in distances over serial and output point cloud of obstacles.

See also pr in roboracing-firmware.


rr_platform/src/ultrasonic_array/ultrasonic_array_to_pcl.cpp
```diff
@@ -0,0 +1,232 @@
...
+int num_sensors;
+int num_points;
+int draw_method;
+int baud_rate;
+
+
+int main(int argc, char** argv) {
+	ros::init(argc, argv, "ultrasonic_array");
+
+	ros::NodeHandle nh;
```
barulicm: Check the formatting in this area, Indentation is inconsistent and it looks like it switches between tabs and spaces a few times.

rr_platform/src/ultrasonic_array/ultrasonic_array_to_pcl.cpp
```diff
@@ -0,0 +1,232 @@
...
+
+    }
+}
+
+vector<double> parseLine(string line) {
+  vector<double> dist;
+  vector<string> strs;
+  boost::split(strs, line, boost::is_any_of(","));
+
+  for (int i = 0; i < strs.size(); i++) {
```
barulicm: This is a good place to use [std::transform](http://en.cppreference.com/w/cpp/algorithm/transform)

rr_platform/src/ultrasonic_array/ultrasonic_array_to_pcl.cpp
```diff
@@ -0,0 +1,232 @@
...
+
+        pcl_ros::transformPointCloud(cloud, cloud, tf_transform_vector[i]);
+
+        //add point cloud to the output cloud
+        compiled_cloud += cloud;
+      }
+    }
+
+
+    //##########################
```
barulicm: I'm personally not a huge fan of big comments to act as separators in code. If you have sections that are logically separate, consider wrapping them in functions. This will both indicate the different sections and give them meaningful names.

rr_platform/src/ultrasonic_array/ultrasonic_array_to_pcl.cpp
```diff
@@ -0,0 +1,232 @@
...
+
+
+    for (int i = 0; i < num_sensors; i++) {
+      pcl::PointCloud<pcl::PointXYZ> cloud;
+
+      if(distances[i] < distance_clip) {
+        pcl::PointXYZ point(distances[i], 0.0, 0.0);
+
+        cloud.push_back(point); //add point direct from Arduino sensor
+        switch (draw_method) {
```
barulicm: Consider moving this out of the while loop. Basically, do the switch to pick a function, store that in a variable, and call the chosen function via that variable rather than redeciding the function to call every iteration.

rr_platform/src/ultrasonic_array/ultrasonic_array_to_pcl.cpp
```diff
@@ -0,0 +1,232 @@
...
+            case 0: break;
+            case 1: drawWall(cloud, point, length, num_points);
+                    break;
+            case 2: drawCircle(cloud, point, radius, num_points);
+                    break;
+            case 3: drawSemiCircle(cloud, point, radius, num_points);
+                    break;
+        }
```
barulicm: Because draw_method is just an int, you should add a `default:` case to this switch.
Note that you can stack case labels, so doing something like `default: case 0: break;` is OK.

rr_platform/launch/ultrasonic_array.launch
```diff
@@ -0,0 +1,15 @@
+<launch>
+  <node pkg="rr_platform" type="ultrasonic_array" name="ultrasonic_array_to_pcl" output="screen" respawn="false">
+    <param name="number_of_sensors" type="int" value="6" />
+    <param name="baud_rate" type="int" value="9600" />
+    <param name="serial_port_name" type="string" value="/dev/ttyUSB0" />
+    <param name="sensor_base_link" type="string" value= "/ultrasonic_array_base" />
+    <param name="sensor_link" type="string" value= "ultrasonic_" />
```
JasonGibson274: feel free to ignore but it might be nice to have prefix somewhere in the name to signal what it is used for later.

barulicm: what prefix are you talking about?

JasonGibson274: this parameter is used as a prefix for the sensor array tf elements.

barulicm: I see what you mean. I like it. 
@btdubs Consider naming this parameter "sensor_link_prefix"

PR #135, opened 2018-02-20
RVIZ Panel for Speedometer and Wheel

mukundSai:
I made a panel for visualizing the speed and angle of the wheel.


rr_rviz_plugins/include/rr_rviz_plugins/SpeedometerPanel.h
```diff
@@ -0,0 +1,71 @@
...
+#include <rr_platform/chassis_state.h>
+#include <rr_platform/steering.h>
+#include <QLabel>
+#include <QtGui>
+#include <QPainter>
+#include <math.h>
+#include <QString>
+
+
+/*
```
JasonGibson274: maintainer preference but I would remove the comments that came with the example panel now that you implemented something.

barulicm: Yes, @mukundSai ^

JasonGibson274:
It also looks like you have not configured git correctly on your machine

rr_rviz_plugins/include/rr_rviz_plugins/SpeedometerPanel.h
```diff
@@ -0,0 +1,53 @@
...
+
+
+
+namespace rr_rviz_plugins {
+
+
+class SpeedometerPanel : public rviz::Panel {
+
+
+
```
chsahit: Just a small note on style, we don't need this much whitespace between class declaration and then starting to populate the class. Ditto throughout the rest of these changes where there are just chunks of whitespace where comments used to be.

ebretl: @chsahit do you still think this needs to be fixed before I approve this PR?

PR #138, opened 2018-02-22
Smoother Steering in Planner

Holben888:
Uses an array of sorted previous "best curve" values to find the median, reducing noise and resulting in smoother motion for the path.


rr_common/src/planner/planner.cpp
```diff
@@ -105,6 +106,18 @@ float aggregateCost(control_vector &controlVector, pcl::KdTreeFLANN<pcl::PointXY
...
 }
 
+/**
+ * Member returning the current bestAngle averaged against the median.
+ * Median is calculated based on previous angles, or PREV_STEERING_ANGLES
+*/
+float steeringAngleAverageAgainstMedian(float bestAngle) {
+    PREV_STEERING_ANGLES_INDEX = (PREV_STEERING_ANGLES_INDEX + 1) % SMOOTHING_ARRAY_SIZE;
+    std::vector<float> sortedSteeringAngles(PREV_STEERING_ANGLES);
+    sort(sortedSteeringAngles.begin(), sortedSteeringAngles.end()); //sort array of previous vectors in ascending order to find median
```
barulicm: Is this the only code that uses the sorted nature of sortedSteeringAngles? If so, consider using [`nth_element`](http://en.cppreference.com/w/cpp/algorithm/nth_element). It does a partial sort on the container: just enough to give you a correct answer for what value would be at position N if the container were fully sorted. 

Note that `nth_element` modifies the container it's given, so you'll still have to copy the values into a temporary container as you've done here.

rr_common/src/planner/planner.h
```diff
@@ -6,6 +6,7 @@
 #include <cstring>
 #include <string>
 #include <random>
+#include <algorithm>
```
barulicm: Try to limit the scope of include statements. Only include headers in the files that directly use them.
In this case, you use the contents of `<algorithm>` in `panner.cpp` but include it in `planner.h`. The side effect here is that any files that include `planner.h` now have to include `<algorithm>` even though they won't be using it directly.

ebretl:
Solid PR. We should note that the math for choosing the proper median element only works for odd-sized buffers.

Issue #132, opened 2018-02-13
Train a model to learn steering angles without overfitting


chsahit:
Note: we want it to learn how to turn, achieving high accuracy from always guessing "straight" is not going to work.

Issue #131, opened 2018-02-13
Collect several minutes of bag data containing images from IARRC sim and steering angles


chsahit:
The expert system would likely be human, using Evan's joystick driver

As we collect this data, we want to drop objects into the background of the track

Issue #130, opened 2018-02-13
Make new IARRC-style tracks with lots of turns.


chsahit:

ebretl:
@chsahit what's the current progress on this?

chsahit:
Assets for a new track exists. Theres a missing dependency preventing the track from actually showing up

Issue #140, opened 2018-03-02
Improve Gazebo Track Textures


barulicm:
The IGVC team's found some documentation that explains how to actually layer textures in Gazebo models!
http://wiki.ogre3d.org/Materials#Alpha_Splatting

We should try to use this to layer different track markings (ie IARRC lines) on top of a more detailed, seamless road texture. This could make it easier to layout different tracks while retaining realistic textures.

For a reference on the general concept of Alpha Splatting, read this: 
https://en.wikipedia.org/wiki/Texture_splatting

JasonGibson274:
This is the method we use for the autorally textures. Let me know if you need help with it when you reach it.

JasonGibson274:
1. write a script that takes in an image (RGB) and outputs the correct alpha image (convert anything that is supposed to be road to full transparency. in rr_gazebo/scripts/texture. It will put the image in rr_gazebo/materials 
2. create a gazebo material that looks like this 

material Gazebo/Texture_blend_[track name]
{
  technique
  {
    pass
    {
      ambient 0.1 0.1 0.1 0.1
      diffuse 0.8 0.8 0.8 1.0
      specular 0.01 0.01 0.01 1.0 2.0

      texture_unit
      {
        [use gazebo texture for road here]
        filtering trilinear
        scale .01 .01 
      }
      texture_unit
      {
        texture textures/[you output image]
        colour_op alpha_blend
      }
      texture_unit
      {
        texture textures/[the image of the track that was drawn]
        colour_op_ex blend_current_alpha src_texture src_current
        alpha_op_ex source1 src_manual src_current 1.0
        scale .01 .01 
      }
    }
  }
}

3. create a model that uses that texture

JasonGibson274:
Other things I think you should condense all the material files into a single material file.

PR #144, opened 2018-04-05
Adds CNN-based planner  to  rr_common

### LINKED ISSUE #143: Python dependencies not included in package.xml
I know this is not on your master branch but I do not want to forget about this. You can add pip dependencies to your package.xml but your CI will fail if you do not install pip onto your docker image. This is in direct reference to keras and h5py

chsahit:
Closes #143 #130 #131 #132


README.md
```diff
@@ -44,8 +44,7 @@ The following files and folders enable our continuous integration system.
    ```
    rosdep install --from-path src --ignore-src -y
    ```
-    
-4. You should now be able to build the project with `catkin_make`.
+5. You should now be able to build the project with `catkin_make`.
```
barulicm: Why did this get changed?

barulicm:
@JasonGibson274 #140 will be separate. This PR just also contains some work @chsahit did a while ago to make some worlds for ML training/testing.

README.md
chsahit: There was a line with install instructions for h5py/keras in the readme that I put in. Then I took it out when I added the dependencies to the package.xml. I'll add it back in since rosdep doesn't yet have the packages.

rr_description/urdf/macaroni.urdf
```diff
@@ -228,13 +228,13 @@
                 </clip>
             </camera>
             <always_on>true</always_on>
-            <update_rate>30</update_rate>
+            <update_rate>10</update_rate>
```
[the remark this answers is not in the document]
ebretl: I was collecting bag files from simulation and was running out of space. I dropped the frame rate to get the breadth of data I wanted without running out of space. The change was meant to be temporary, good catch

chsahit:
@JasonGibson274 I don't recall if the rectifier was running on the images we collected, maybe @ebretl knows?

rr_common/src/nn_planner/nn_planner.py
```diff
@@ -0,0 +1,77 @@
...
+    rospy.init_node('nn_planner')
+
+    path_to_model = os.path.join(os.path.dirname(__file__), rospy.get_param('~model_path'))
+    model = load_model(path_to_model)
+    input_shape = model.get_layer(index=0).get_input_shape_at(0)[1:]
+
+
+    publisher = rospy.Publisher('/steering', Steering, queue_size=1)
+    speed_publisher = rospy.Publisher('/speed', Speed, queue_size=1)
+    rospy.Subscriber('/camera/image_rect', Image, plan_rect)
```
dpattison3: You may not have a problem with this if the network runs fast enough, but only setting the queue size on image subscribers doesn't  always work quite how you'd expect. The default buffer size is too small for images and if the pipeline takes too long, there will be a significant lag (even though you'd expect it to only use the most recent image). See IGVC's subscriber as an example https://github.com/RoboJackets/igvc-software/blob/master/igvc/src/cnn/cnn.py#L75

PR #145, opened 2018-05-25
Updates README

ebretl:


barulicm:
Can we upload that image to the wiki repo for this repo like the old image was? That way our README's not depending on someone's Google Photos URL.

barulicm:
Also, "mission code," so fancy :smile: :+1:

ebretl:
@barulicm 😎. Which repo hosts files like that?

JasonGibson274:
This is probably an open question for both IGVC and RR repos but who are the installation instructions for? yours would be fine for anyone who is familiar with ROS but no new member would be able to follow it IMO. This is what IGVC has https://wiki.robojackets.org/IGVC_Software_Installation_Instructions

IGVC is getting more complicated with additional external to ROS dependencies. I want to generate something similar to AR at a goal of what it should start looking like.

barulicm:
@ebretl fun fact, the GitHub wikis are actually git repos under the hood.
The one for this repo is at https://github.com/RoboJackets/roboracing-software.wiki.git, which you can find when you go to the wiki tab on any repo, then look at the right hand side of the page.

barulicm:
@JasonGibson274 that's a good question, though I don't know that IGVC's instructions linked are actually that much more useful than ours. It looks like you've just done the same thing, but then copied the ROS install instructions into your own. I'd like to avoid that redundancy.
In general, I think the vast majority of new members won't be able to follow any written set of instructions without supervision, so this is mostly just for people already familiar with ROS. They might be above average rookies, or someone external to RJ all together. In my experience, we have to hold most peoples hands through those first steps in person regardless of how much we write down.

JasonGibson274:
I fully agree that the IGVC instructions need to be updated a lot and are not useful right now. That answer of target audience works for me.

PR #155, opened 2018-07-23
IARRC 2018 Competition Merge

barulicm:
All of the IARRC 2018 competition changes. Let's review them! Yay! :tada:


iarrc/src/color_detector/color_detector.cpp
```diff
@@ -40,41 +44,45 @@ namespace iarrc {
 
         const Mat &frameBGR = cv_ptr->image;
         Mat frameBlurred;
-        GaussianBlur(frameBGR, frameBlurred, Size{7,7}, 0);
+        GaussianBlur(frameBGR, frameBlurred, Size{0,0}, 7);
```
ebretl: Just a note/FYI for OpenCV users. The last parameter of GaussianBlur is the standard deviation of the distribution that populates the linear filter. So before this change, there was not gaussian filtering going on. The size argument specifies the shape of the linear filter, and if it's (0,0) then the filter will be automatically sized based on the standard deviation.

rr_platform/launch/pointgrey_camera.launch
```diff
@@ -21,30 +22,35 @@
...
     <param name="auto_exposure" value="false"/>
-    <param name="exposure" value="1.25"/>
+    <param name="exposure" value="0.001"/>
     <param name="auto_brightness" value="false"/>
     <param name="brightness" value="0"/>
-    <param name="auto_white_balance" value="false"/>
+    <param name="auto_white_balance" value="true"/>
     <param name="auto_shutter" value="false"/>
-    <param name="auto_gain" value="false"/>
+    <param name="auto_gain" value="true"/>
```
ebretl: We found that changing these values changed neither the image quality nor the values in rqt_reconfigure. We used reconfigure at competition to adjust the camera params

PR #158, opened 2018-07-27
IARRC 2018 Competition Merge (2nd Attempt)

barulicm:
Resetting so we can go through the full review process.

As of July 26, @ebretl and @luisp23 had approved the changes included (via the first PR attempt).


barulicm:
With regards to the IARRC launch files moved into an 'old' folder, I think we should just remove them completely. We can always dig them back from git history if we need to.

rr_platform/msg/axes.msg
```diff
@@ -0,0 +1,5 @@
+# message for describing the axes of the car in degrees
```
barulicm: Why are we using a custom message instead of using just pulling the orientation from the standard sensor_msgs/Imu message the node is already publishing?

rr_platform/src/imu/imu.cpp
```diff
@@ -0,0 +1,139 @@
...
+    while(ros::ok()) {
+        ros::spinOnce();
+
+        auto line_in = serial_port.ReadLine();
+
+        if(!line_in.empty()) {
+
+            parse_message(line_in, imu_msg, mag_msg, axes_msg, imu_ready, mag_ready, axes_ready);
+
+            if(imu_ready) {
```
barulicm: This whole ready flag system is very unnatural for ROS. 
I'd recommend breaking up the different parsing blocks into their own functions. Each one returns a pair with the message and a success bool. Then this main loop can simply ask each function to look for its message type in the line, and if it finds it, this loop will publish it.
This idea will also clean up very nicely when we can use C++17's if statements with initializers.

util/neural_nets/params.py
```diff
@@ -1 +1,26 @@
 input_shape = (48, 128, 3) # rows, cols, channels
+
```
barulicm: Can we move these neural_net scripts to an actual ros package? They don't have to use ROS code, but that will make them usable via `rosrun` so you could run them from anywhere on your machine.
Could call it rr_nn or rr_ml or whatever makes sense.

rr_platform/msg/axes.msg
luisp23: The new IMU computes Eueler angles and outputs with the rest of the data. The sensor_msgs/Imu message did not have a field for this so I made this custom message.

iarrc/launch/nn_planner_circuit.launch
```diff
@@ -0,0 +1,11 @@
+<launch>
+    <include file="$(find rr_common)/launch/nn_planner.launch">
+        <arg name="image_topic" value="camera/image_color_rect_flipped"/>
```
chsahit: are we using the old /image_rect topic? if not we might want to remap image_color_rect_flipped into that. The latter is unnecessarily verbose imo

barulicm: image_color_rect_flipped just stems from the ROS conventions on building names by appending the operations done.
We'll be able to get rid of the _flipped part anyway as we're moving the camera to the other side of the plate it's mounted to, which will put it back to being upright.
Whether we keep the _color in there is up to us. It distinguishes it from the monochromatic topic that [image_proc/debayer](http://wiki.ros.org/image_proc) publishes.

rr_platform/launch/imu.launch
```diff
@@ -1,5 +1,6 @@
+<?xml version="1.0" encoding="utf-8" ?>
 <launch>
     <node name="imu" pkg="rr_platform" type="imu" output="screen">
-        <param name="serial_port" value="/dev/ttyUSB0"/>
+        <param name="serial_port" value="/dev/ttyACM0"/>
```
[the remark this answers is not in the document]
barulicm: Issue added

rr_platform/msg/axes.msg
barulicm: ok, sounds reasonable :+1:

util/neural_nets/params.py
barulicm: @ebretl thoughts on this?

ebretl: Issue created

PR #172, opened 2018-09-19
Instructions for installing needed packages from source

ebretl:


README.md
```diff
@@ -44,13 +44,19 @@ The following files and folders enable our continuous integration system.
    ```sh
    rosdep install --from-path src --ignore-src -y
    ```
+
+	- Some packages may have to be installed from source. For example, if realsesnse2_camera cannot be found, then search the error and find the ROS wiki page http://wiki.ros.org/realsense2_camera. Find the link to the source git repo and clone that into catkin_ws/src.
```
barulicm: Since we know the two repos that need this step and this is an exception to the normal rosdep workflow, can we just add instructions that say something like "clone ____ and ___ to src" like we've done with Keras.

ebretl: Fair

barulicm:
Is there a reason you're using `sudo pip install` instead of `pip install --user` I thought sudo-ing pip was discouraged?

ebretl:
No, to be honest I've always installed Python packages with sudo. You're right, it is a security risk. What's kind of stupid is that if you have already installed it with sudo, then it looks like installing again with --user will download it a second time.

Issue #156, opened 2018-07-26
Tool for converting bag files to video files


barulicm:
PR could use some cool FPV footage from our cars. We can generate this pretty easily from our bag files, we just need a short tool to do it.

Useful resources for this task:

- [ROS BAG C++ API](http://wiki.ros.org/rosbag/Code%20API)
- [OpenCV: How to write a video](https://docs.opencv.org/2.4/doc/tutorials/highgui/video-write/video-write.html)
- [OpenCV: VideoWriter documentation](https://docs.opencv.org/2.4/modules/highgui/doc/reading_and_writing_images_and_video.html?highlight=videowriter#videowriter)

PR #187, opened 2018-10-23
Bag to video python script

dmulloy2:
Pretty simple python script to turn the images we store in the bag files to a video for PR and such. Requires the rosbag and opencv python packages (which might get installed automatically by ros, don't 100% remember). Video settings can be changed with command line flags (see script for details)

See #156


ebretl:
Yes, I believe that ROS is shipped with opencv-python so we should be good on dependencies

PR #189, opened 2018-11-01
Demostration

Daniel-Martin576:
This demonstration mode moves the car forwards or backwards depending on the average distance between itself and the objects in its bounded line of sight. Currently, it will move backwards until it passes the goal distance and then move forwards. In the future, a gradually decreasing backwards/forward movement could be added to remove this action.


rr_common/src/follower/follower.cpp
```diff
@@ -0,0 +1,100 @@
...
+    pcl_conversions::toPCL(*map, pcl_pc2);
+    pcl::PointCloud<pcl::PointXYZ>::Ptr cloud(new pcl::PointCloud <pcl::PointXYZ>);
+    pcl::fromPCLPointCloud2(pcl_pc2, *cloud);
+    rr_platform::speedPtr speedMSG(new rr_platform::speed);
+    rr_platform::steeringPtr steerMSG(new rr_platform::steering);
+    if (cloud->empty()) {
+        ROS_WARN("environment map pointcloud is empty");
+        speedMSG->speed = 0;
+        steerMSG->angle = 0;
+        steerMSG->header.stamp = ros::Time::now();
```
JasonGibson274: I am not sure how this timestamp is used, but you might want to check this makes semantic sense

rr_common/src/follower/follower.h
```diff
@@ -0,0 +1,25 @@
...
+#include <pcl/point_cloud.h>
+#include <pcl/PCLPointCloud2.h>
+#include <pcl_conversions/pcl_conversions.h>
+#include <pcl/filters/passthrough.h>
+#include <string>
+#include <rr_platform/speed.h>
+#include <rr_platform/steering.h>
+#include "flann/flann.hpp"
+
+float MIN_X;
```
JasonGibson274: I would consider a better name, something that represents obstacles in this since I was initially unsure what this was used for without reading through the code.

ebretl:
I thought we were going to be able to test this fall semester. Probably not worth holding the PR open any longer. We can test this once Sedani can drive backwards.

PR #192, opened 2018-11-06
Restructures avc, iarrc, ml packages

### LINKED ISSUE #162: Create ROS package for machine learning code
Our data-processing and model-training scripts do not live in a ROS package, and so they cannot be executed as easily as our other code. Create a package rr_ml to hold these scripts and future ones of the same variety

ebretl:
Fixes #162 

Refactors "avc" and "iarrc" packages and namespaces to "rr_avc" and "rr_iarrc" respectively.

Also creates a new "rr_ml" package which consolidates our ML code into one place. rr_ml is structured for Python following the "scripts" section of http://docs.ros.org/melodic/api/catkin/html/howto/format1/installing_python.html. It does not install any modules via catkin.


rr_ml/.gitignore
```diff
@@ -0,0 +1 @@
+exported_data/*
```
barulicm: Any reason this rule isn't just part of the top-level gitignore?

rr_ml/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
+cmake_minimum_required(VERSION 2.8.3)
+project(rr_ml)
+
+list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
```
barulicm: I don't see this `cmake` directory in the rr_ml package.

PR #201, opened 2019-01-22
Fix for sedani not publishing to chassis_state

btdubs:
We accidentally removed the ability for the relay node to publish Sedani's data at competition. This adds it back.


ebretl:
Do we know that publishData parses the serial data in the right order?

btdubs:
Yes, firmware sends data as State, Speed, Steering Angle which is the ordering publishData expects.

PR #202, opened 2019-01-24
fixes sedani chassis state interpretation

ebretl:


btdubs:
Isn't estopped state 0? And what is the distinction between disabled and estopped? Disabled in the firmware is estopped I thought, but here is looks to mean either manual mode or regular.

jspall0327:
Sorry for the confusion. State 0 (disabled) is equivalent to e-stop. We just aren't really "e-stopping" in the industry sense of the term, so that is why we clarified to disabled. The car can be disabled in either manual or auto though. State 2 (manual) is the same manual as before.

jspall0327:
Also kind of nit-picky, but auto stands for autonomous not automatic. Might want to change that message definition.

PR #204, opened 2019-02-11
Laplacian Code Implementation

Daniel-Martin576:


rr_iarrc/src/laplacian_line_detection/laplacian_line_detector.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+cv_bridge::CvImagePtr cv_ptr;
+ros::Publisher pub, pub1, pub2, pub3;
+
+cv::Mat kernel(int, int);
+cv::Mat fillColorLines(cv::Mat, cv::Mat);
+void cutEnvironment(cv::Mat);
+cv::Mat cutSmall(cv::Mat, int);
+void publishMessage(ros::Publisher, Mat, std::string);
+Mat overlayBinaryGreen(Mat, Mat);
```
ebretl: I like that you're using forward declarations. Helps give people an overview of what to expect in this file

PR #206, opened 2019-02-11
Pointcloud conversion

### LINKED ISSUE #197: Image to pointcloud converter pipeline needs some TLC
- [ ] everything we use from rr_platform/include/avc/constants.hpp should be specified via roslaunch params/args and the constants file should be deleted
- [ ] image perspective warping and conversion to a pointcloud should happen within one node. This can make it a nodelet like iarrc's color_detector, but it does not have to be. My recommendation is to start with a normal node and if the latency is an issue, we can optimize it later
- [ ] the camera pose calibration code should be moved to a new file, say, calibrate_camera_pose.cpp. It should compile but don't worry about verifying that it works. It's only around for reference if we need to do something like that again. Consider making a folder like rr_platform/src/deprecated or something like that

ebretl:
- changes pcl_converter to use single-channel images, not sure why it was 3-channel before
- removes all camera pose calibration code (I wrote it freshman year but it was never needed or even tested afaik)

Fixes #197


rr_common/src/image_pcl_converter/image_pcl_converter.cpp
```diff
@@ -17,34 +17,38 @@
...
     }
 
-    pcl::PointCloud<pcl::PointXYZ>::Ptr cloud(new pcl::PointCloud<pcl::PointXYZ>);
-    Mat transformed = cv_ptr->image;
+    cloud->clear();
+    const cv::Mat& in_image = cv_ptr->image;
+
+    // get outline
+    cv::Mat transformed(in_image.size(), CV_8UC1);
+    cv::Canny(in_image, transformed, 10, 0);
```
btdubs: Nitpicking but since we have a binary image, canny takes more steps than needed. A simple Laplace should work just as well and do less work. Laplace will output gradients and not just an binary edge image, so the below check will need to account for negative numbers as well: `row[c] != 0;`

ebretl: Changing Canny to Laplacian does seem to speed up the inner pointcloud conversion loop, but it already took less than a millisecond anyway :man_shrugging:

rr_common/src/image_pcl_converter/image_pcl_converter.cpp
```diff
@@ -38,21 +38,20 @@ void transformedImageCB(const sensor_msgs::ImageConstPtr& msg, const string& top
...
-        uchar* row = transformed.ptr<uchar>(r);
+        auto* row = transformed.ptr<int16_t>(r);
         for(int c = 0; c < transformed.cols; c++) {
-            if(row[c] > 0) {
-
+            if(row[c] != 0) {
                 pcl::PointXYZ point;
-                point.y = -1 * (c - transformed.cols / 2.0f) / pxPerMeter;
-                point.x = (transformed.rows - r) / pxPerMeter;
+                point.y = static_cast<float>((transformed.cols - c / 2.0f) / pxPerMeter);
```
btdubs: I think you forgot to move the divisor so this should be `((transformed.cols / 2.0f) - c) / pxPerMeter`

ebretl: Good catch, that would have been annoying to find

PR #213, opened 2019-02-11
fixes wrong param in test_scanToPointCloud.test

### LINKED ISSUE #210: Failed rostests don't trigger CI failure
A test failed on master but Circle says everything is fine and passing: https://circleci.com/gh/RoboJackets/roboracing-software/840

The cause of the test failure is simply a bad value in test_scanToPoincloud.test that was causing us to test the wrong thing (the node is working as intended)

ebretl:
This is causing the failed test that CircleCI has been missing for some reason (issue is open for that separate problem)

Edit: also fixes the issue where Circle does not fail on a failed test case. The solution was pretty easy so it's included here. Fixes #210

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -80,7 +80,7 @@ jobs:
       - <<: *install_deps
       - <<: *load_workspace
       - <<: *load_test_cache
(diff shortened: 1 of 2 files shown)


barulicm:
@JasonGibson274 The IGVC circle config will need this fix as well.

PR #217, opened 2019-02-12
collects miscellaneous changes from iarrc-integration

ebretl:
- documentation and bugfixes in launch files
 - updates cpp version to c++17 in all cmakelists


rr_avc/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(rr_avc)
 
-set(CMAKE_CXX_FLAGS "-std=c++11 ${CMAKE_CXX_FLAGS}")
+set(CMAKE_CXX_FLAGS "-std=c++17 ${CMAKE_CXX_FLAGS}")
```
barulicm: Consider using `set(CMAKE_CXX_STANDARD 17)`

ebretl: Other than looking cleaner, is there a functional difference between the two?

barulicm: The [CXX_STANDARD](
https://cmake.org/cmake/help/v3.6/prop_tgt/CXX_STANDARD.html) documentation lays out the specific behavior. 

For this project, as we do not do any cross compiler or cross platform development, it wouldn’t have any functional effect. It is simply the better practice way of doing this as directly editing the cflags is a risky habit to get in. (It’s easy to do it wrong and limits cross-platform support).

PR #216, opened 2019-02-12
Neural Network segmentation

ebretl:
This code accomplishes:
 - labeling images or folders of images with polygons (LabelMe)
 - training a small semantic segmentation neural network
 - a line detector backed by a pre-trained neural network
 - all labeling and training scripts are run via roslaunch

PR #218, opened 2019-02-12
test launch files for image-to-planner pipeline

ebretl:
These launch files depends on the neural network, so #216 must be merged first

### DIFF
--- a/rr_iarrc/launch/line_detector_segnet.launch
+++ b/rr_iarrc/launch/line_detector_segnet.launch
@@ -1,0 +1,7 @@
+<launch>
+    <include file="$(find rr_ml)/launch/run_seg.launch">
+        <arg name="image_topic" value="/camera/image_color_rect"/>
+        <arg name="detection_topic" value="/lines_detection_img"/>
+        <arg name="model" value="iarrc_lines"/>
+    </include>
+</launch>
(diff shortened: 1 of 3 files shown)


rr_iarrc/launch/line_detector_segnet.launch
```diff
@@ -0,0 +1,7 @@
+<launch>
+    <include file="$(find rr_ml)/launch/run_seg.launch">
+        <arg name="image_topic" value="/camera/image_color_rect"/>
+        <arg name="detection_topic" value="/lines_detection_img"/>
+        <arg name="model" value="iarrc_lines"/>
```
JasonGibson274: Does this assume a specific path?

rr_iarrc/launch/test_segnet_and_planner.launch
```diff
@@ -0,0 +1,8 @@
+<launch>
+    <include file="$(dirname)/line_detector_segnet.launch"/>
```
JasonGibson274: why $(dirname) rather than package relative?

rr_iarrc/launch/line_detector_segnet.launch
ebretl: Kind of. There is a specific folder in which it looks for iarrc_lines.h5. That is set in run_seg.launch

rr_iarrc/launch/test_segnet_and_planner.launch
ebretl: Purely because it's shorter. Is it best practice to use package-relative even when the included file is in the same folder?

PR #226, opened 2019-03-01
automates setting gazebo physics settings

ebretl:
Default settings are 1000 updates per second at 1ms per update. Now it runs with significantly less CPU usage and still seems to simulate motion on macaroni_avc.launch equally as well as before

### DIFF
--- a/rr_gazebo/scripts/spawn_macaroni_platform.sh
+++ b/rr_gazebo/scripts/spawn_macaroni_platform.sh
@@ -3,3 +3,7 @@
 sleep 3
 
 roslaunch rr_gazebo spawn_macaroni_platform.launch x:=$1 y:=$2 heading:=$3
+
+# set gazebo physics step size so that it uses less CPU resources
+sleep 2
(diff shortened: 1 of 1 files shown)


rr_gazebo/scripts/spawn_macaroni_platform.sh
```diff
@@ -3,3 +3,7 @@
 sleep 3
 
 roslaunch rr_gazebo spawn_macaroni_platform.launch x:=$1 y:=$2 heading:=$3
+
+# set gazebo physics step size so that it uses less CPU resources
+sleep 2
+gz physics --step-size 0.0025 --update-rate 400 --iters 30
```
JasonGibson274: You can do this directly in the launch file.

ebretl: I tried that but then the settings have to be set in all 4 world models that we have. That's not such a bad thing but I'm not sure which is better

PR #227, opened 2019-03-11
Faster Laplacian Line Detection

Daniel-Martin576:


rr_iarrc/launch/laplacian_line_detector.launch
```diff
@@ -1,28 +1,21 @@
 <launch>
     <node name="laplacian_line_detector" pkg="rr_iarrc" type="laplacian_line_detector" output="screen">
         <!--Minimum Area to Keep-->
-        <param name="perfect_lines_min_cut" type="int" value="350" />
+        <param name="perfect_lines_min_cut" type="int" value="50" />
```
ebretl: iirc lowering the min cut before changed the behavior a lot. Is that correct, and if so, I'm curious which change(s) now make 50 a good value for the min cut?

rr_iarrc/src/laplacian_line_detection/laplacian_line_detector.cpp
```diff
@@ -12,53 +12,58 @@
 #include <stdio.h>
 
 using namespace cv;
```
ebretl: Separate from this PR, it like to see us remove all "using namespace cv" invocations at some point in the future.

rr_iarrc/src/laplacian_line_detection/laplacian_line_detector.cpp
```diff
@@ -70,14 +75,15 @@ int main(int argc, char** argv) {
...
     nhp.param("blockWheels_height", blockWheels_height, 200);
     nhp.param("blockBumper_height", blockBumper_height, 200);
 
     nhp.param("subscription_node", subscription_node, std::string("/camera/image_color_rect"));
 
-    pub = nh.advertise<sensor_msgs::Image>("/Combined_lines", 1); //test publish of image
+    pub = nh.advertise<sensor_msgs::Image>("/lines_detection_img", 1); //test publish of image
     pub1 = nh.advertise<sensor_msgs::Image>("/Adpt_Thres_lines", 1);
     pub2 = nh.advertise<sensor_msgs::Image>("/Laplacian_lines", 1);
     pub3 = nh.advertise<sensor_msgs::Image>("/Colored_combined_lines", 1);
```
ebretl: As a future task, consider improving this node by making it a nodelet and using image_transport for publishing images. Ref http://www.clearpathrobotics.com/assets/guides/ros/Nodelet%20Everything.html

rr_iarrc/src/laplacian_line_detection/laplacian_line_detector.cpp
```diff
@@ -12,53 +12,58 @@
...
-//	publish images
     publishMessage(pub, fill, "mono8");
-    publishMessage(pub1, cut, "mono8");
+    publishMessage(pub1, thres, "mono8");
     publishMessage(pub2, lapl, "mono8");
-    publishMessage(pub3, green_lines, "bgr8");
+
+//    Uncomment for Green Overlay
+//    Mat green_lines = overlayBinaryGreen(frame, fill);
+//    publishMessage(pub3, green_lines, "bgr8");
```
ebretl: This might be a good place for checking pub3.getNumSubscribers() and publishing if > 0

rr_iarrc/src/laplacian_line_detection/laplacian_line_detector.cpp
```diff
@@ -12,53 +12,58 @@
...
-    GaussianBlur(frame, frame_blur, Size(5,5), 0, 0, BORDER_DEFAULT);
+    GaussianBlur(frame, frame_blur, Size(5,5), 0);
     cv::cvtColor(frame_blur, frame_gray, cv::COLOR_BGR2GRAY);
 
     Mat thres;
     adaptiveThreshold(frame_gray, thres, 255, ADAPTIVE_THRESH_GAUSSIAN_C, THRESH_BINARY, 5, -1);
-    cutEnvironment(thres);
-    Mat cut = cutSmall(thres, perfect_lines_min_cut);   //Make sure only Lines
+    cutEnvironment(thres, 0);
+    erode(thres, thres, kernel(3,1));
```
ebretl: Harping on importing the cv namespace again: my gut reaction here was to consult the OpenCV docs on cutEnvironment, but those don't exist because it's ours

rr_iarrc/src/laplacian_line_detection/laplacian_line_detector.cpp
```diff
@@ -95,45 +101,43 @@ cv::Mat kernel(int x, int y) {
 cv::Mat fillColorLines(cv::Mat lines, cv::Mat color_found) {
```
ebretl: I __think__ OpenCV will try its best to remove unnecessary copying of data from within its Mat class. However, best practice dictates using const references (const LargeObject& my_param) to minimize copying. Often times, writing fast C++ code is a matter of avoiding copying data in memory.

PR #228, opened 2019-03-14
Testday changes

ebretl:
General fixes and improvements from the Feb 24 test day, pulled out of the iarrc-integration branch

### DIFF
--- a/rr_description/urdf/sedani.urdf
+++ b/rr_description/urdf/sedani.urdf
@@ -321,7 +321,7 @@
     <joint name="base_footprint_to_chassis" type="fixed">
         <parent link="base_footprint"/>
         <child link="chassis"/>
-        <origin xyz="0 0 0.036" rpy="0 0 0"/>
+        <origin xyz="0 0 0.079" rpy="0 0 0"/>
     </joint>
 
     <joint name="chassis_to_back_axle" type="fixed">
@@ -374,7 +374,7 @@
     <joint name="camera_joint" type="fixed">
         <parent link="chassis"/>
         <child link="camera"/>
-        <origin xyz="0.28 0 0.08" rpy="0 -0.04 0"/>
+        <origin xyz="0.28 0 0.1" rpy="0 -0.07 0"/>
     </joint>
 
     <joint name="camera_optical_joint" type="fixed">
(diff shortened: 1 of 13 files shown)


rr_iarrc/launch/lines_pcl_converter.launch
```diff
@@ -1,9 +1,9 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <launch>
     <node pkg="rr_platform" type="image_transform" name="image_transform" output="screen">
-        <param name="transform_topics" value="/lines_detection_img"/>
+        <param name="transform_topics" value="/lines_detection_img /cone_detection_img"/>
```
btdubs: This should stay `/lines_detection_img `so we don't get confused later since the it is the line converter launch file.

btdubs: And same later when this was done.

rr_iarrc/src/laplacian_line_detection/laplacian_line_detector.cpp
```diff
@@ -77,7 +77,7 @@ int main(int argc, char** argv) {
 
     nhp.param("subscription_node", subscription_node, std::string("/camera/image_color_rect"));
 
-    pub = nh.advertise<sensor_msgs::Image>("/Combined_lines", 1); //test publish of image
+    pub = nh.advertise<sensor_msgs::Image>("/lines_detection_img", 1); //test publish of image
```
btdubs: We may want to come up with a better naming patter for our line detection topics. Maybe `/lines/<thing>` since we have a couple secondary debug ones to go along with the final lines output.

btdubs: Also remove the comment since that is the final image. Pretty sure that is from people copying and pasting from my example/test image node haha.

rr_description/urdf/sedani.urdf
```diff
@@ -321,7 +321,7 @@
     <joint name="base_footprint_to_chassis" type="fixed">
         <parent link="base_footprint"/>
         <child link="chassis"/>
-        <origin xyz="0 0 0.036" rpy="0 0 0"/>
+        <origin xyz="0 0 0.079" rpy="0 0 0"/>
```
btdubs: Why is this value changing? Did we remeasure? I know with the camera joint we found certain values that made the transformation work better but this seems like it will affect more than the camera.

ebretl: Yeah sedani.urdf used to be a verbatim copy of macaroni.urdf and so far things have been changing only as needed. This is the real height of Sedani's back axle. I think the lidar joint has been updated, but the IMU joint has not

PR #229, opened 2019-03-14
Fixes some configuration settings that used to be dumb

ebretl:
- compile with release flags (this enables compiler optimizations, which apparently have been turned off for the entire history of this team)
 - sets the camera height in URDF to the actual height on the robot
 - sets the camera tf frame in pointgrey_camera.launch so that we can visualize the camera in Rviz


ebretl:
I replaced setting the flags in CMakeLists with a script people can use if they want. We should definitely build for Release on Sedani IMO

PR #230, opened 2019-03-14
Camera geometry library

ebretl:
- CameraGeometry.h and .cpp: Implements a library to load features of the camera such its pose and field of view. It also provides a function to project a pixel location onto the ground (commonly called backward projection). The capabilities of this library could be extended in the future, for example 1) implementing forward projection and 2) using XYZRGB colored points in the backward projection. Neither of these are hard, I just wanted to merge sooner rather than later.
- angle_utils.hpp: header-only "library" for common operations on angles that come up in various places (idk how many times I've written some version of angle_difference for different projects)
- pointcloud_projector.cpp: An alternative to our existing image-to-pointcloud pipeline which is much easier to read and understand, using the CameraGeometry library. The resulting pointclouds are not significantly different than image_transform + pcl_converter, but they are not the same either. Both are light on CPU usage. The main advantages of pointcloud_projector are (IMO) readability and the fact that it's all wrapped in a single nodelet rather than spread over multiple nodes. Use lines_projector.launch to run this nodelet.


rr_platform/src/camera/CameraGeometry.cpp
```diff
@@ -0,0 +1,146 @@
...
+  geometry_msgs::PoseStamped ps_dst_base;  // destination pose, camera frame origin from base_footprint
+
+  ps_src_cam.pose.orientation.w = 1.0;  // set orientation in x-axis direction
+  ps_src_cam.header.frame_id = camera_link_name;
+
+  tf::TransformListener listener;
+  bool success = false;
+  while (!success) {
+    try {
+      listener.waitForTransform("base_footprint", camera_link_name, ros::Time(0), ros::Duration(timeout));
```
ebretl: This never returns false. Either remove the loop or implement the timeout as part of the loop

PR #235, opened 2019-03-24
Adds basic rosbag recording functionality from E-Stop remote

btdubs:
Press the "D" (bottom right) button on E-Stop Remote to begin recording a rosbag AND un-estop if e-stopped. Pressing any other button (like e-stopping) will stop the recording. This can be changed to be smarter once Sedani has leds that can indicate recording status or has a new remote.
 
Requires the merge of same named branch for firmware first because this edits the motor relay node. [https://github.com/RoboJackets/roboracing-firmware/pull/23](https://github.com/RoboJackets/roboracing-firmware/pull/23)


rr_common/launch/rosbag_remote_recorder.launch
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<launch>
+    <node pkg="rr_common" type="rosbag_remote_recorder" name="rosbag_remote_recorder" output="screen" respawn="true">
+        <param name="topics_to_record" type="string" value="/camera/image_color_rect"/> <!--Seperate multiple topics with a space -->
+	<param name="bag_name_prefix" type="string" value="test"/> 
+	<param name="absolute_folder_path" type="string" value="/home/brian/bag_files/"/>
```
ebretl: Can we make this a relative folder? Also indentation is wonky

btdubs:
Testing this on the car, it starts but the recorder seems to block so the node never gets a chance to shut itself down to end the recording. I will need to look more into this or find a different way.

btdubs:
Note to self: Possible solution to try is call `ros::shutdown()` within the chassis_state subscriber callback  instead of the main method since the recorder blocks.

btdubs:
Above idea to call shutdown in the callback fixes the blocking problem. Rosbag recorder now starts and stops as expected.

rr_common/src/rosbag_remote_recorder/rosbag_remote_recorder.cpp
```diff
@@ -0,0 +1,80 @@
...
+auto pred = [&](const std::string& key) ->bool
+{
+    return key.empty();
+};
+
+void chassisStateCallback(const rr_platform::chassis_state::ConstPtr &chassis_msg) {
+    startRecording = chassis_msg->record_bag;
+    if (!startRecording && begunRecording) {
+        ROS_INFO_STREAM("Stopping bag recording");
+        ros::shutdown(); //only way to stop the recording, but we respawn!
```
ebretl: Does shutdown() kill just this node, or all nodes?

PR #236, opened 2019-03-31
Renamed cone detection and laplacian nodes and Documented code

Daniel-Martin576:
I renamed the cone detection and laplacian nodes based on the given name recommendations and I removed the cv namespace from two of the techniques (laplacian and cone_bottom). I haven't converted them into nodelet yet, but I hope to do so in the coming weeks. Cone_detection_height is still not optimized since the image is being analyzed at normal size (without resize). The resizing increases the speed significantly; however, I fear the pixel height of the cones will be less accurate in doing so.


rr_iarrc/src/cone_detection/cone_detection_height.cpp
```diff
@@ -0,0 +1,336 @@
...
+#include <algorithm>
+#include <pcl_ros/point_cloud.h>
+#include <iostream>
+#include <stdlib.h>
+#include <stdio.h>
+#include <sensor_msgs/CameraInfo.h>
+
+using namespace cv;
+using namespace std;
+using namespace ros;
```
ebretl: Too many namespaces

rr_iarrc/src/cone_detection/cone_detection_bottom.cpp
```diff
@@ -0,0 +1,133 @@
...
+    int originalWidth = frame.cols;
+    cv::resize(frame, frame, cv::Size(400, 400));
+
+    //Get Orange-HSV Cones
+    cv::Mat hsv_frame, orange_found, orange_found_rot, nonZeroCoordinates, green_lines;
+    cv::cvtColor(frame, hsv_frame, cv::COLOR_BGR2HSV);
+    cv::inRange(hsv_frame, cv::Scalar(low_H, low_S, low_V), cv::Scalar(high_H, 255, 255), orange_found);
+    cutEnvironment(orange_found);
+
+    //Rotates Image because findNonZero() gives points already sorted by Y-coordinate
```
btdubs: I am confused why you rotate this image or the `last` variable. I suspect you may be expected behavior from `cv::findNonZero()` that is not guaranteed (giving points from top to bottom). If that is the case, just make function to iterate from the bottom of each column until you hit a white pixel, then add that to your image and continue.

rr_iarrc/launch/laplacian_line_detector.launch
```diff
@@ -3,15 +3,18 @@
...
         <param name="blockSky_height" type="int" value="95" />
-        <param name="blockWheels_height" type="int" value="100" />
-        <param name="blockBumper_height" type="int" value="50" />
+        <param name="blockWheels_height" type="int" value="70" />
+        <param name="blockBumper_height" type="int" value="0" />
 
         <param name="subscription_node" type="string" value="/camera/image_color_rect"/>
     </node>
+
+    <include file="$(find rr_platform)/launch/sedani_state_publisher.launch"/>
```
btdubs: The platform or iarrc competition launch files can handle launching the state publisher and the image transformer. We don't want to link this to a node as that makes it less reusable, so remove these extra launch files.

rr_iarrc/launch/cone_detection_height.launch
```diff
@@ -0,0 +1,28 @@
...
+        <param name="orange_high_H" type="int" value="20" />
+        <param name="orange_low_S" type="int" value="150" />
+        <param name="orange_low_V" type="int" value="140" />
+
+        <param name="canny_cut_min_threshold" type="int" value="60" />
+        <param name="percent_max_distance_transform" type="double" value="0.7" />
+        <param name="minimum_area_cut" type="double" value="30" />
+
+        <!--Formula: F = PxHeight * RealDist / RealHeight-->
+        <param name="camera_focal_length_y" type="double" value="611.46" />
```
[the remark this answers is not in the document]
btdubs: This is the same in the other launch files as well.

rr_iarrc/src/cone_detection/cone_detection_height.cpp
```diff
@@ -0,0 +1,336 @@
...
+#include <boost/algorithm/string.hpp>
+#include <boost/math/constants/constants.hpp>
+#include <algorithm>
+#include <pcl_ros/point_cloud.h>
+#include <iostream>
+#include <stdlib.h>
+#include <stdio.h>
+#include <sensor_msgs/CameraInfo.h>
+
+using namespace cv;
```
btdubs: We would prefer using minimal namespaces. Std is fine, cv and ros should probably not be used.

rr_iarrc/src/cone_detection/cone_detection_height.cpp
```diff
@@ -0,0 +1,336 @@
...
+    return contours_color;
+}
+
+
+cv::Mat cutBodies(cv::Mat frame_cut, cv::Mat orange_found) {
+    Mat frame_gray, detected_edges, detected_bodies;
+
+    GaussianBlur(frame_cut, frame_cut, Size(3,3), 0, 0, BORDER_DEFAULT );
+    cv::cvtColor(frame_cut, frame_gray, cv::COLOR_BGR2GRAY );
+    bitwise_and(frame_gray, orange_found, frame_gray);
```
btdubs: This feels a confusing way to remove the environment from the orange found image. May want to make a note that is why you do it rather than using `cutEnvironment() `.

rr_iarrc/launch/laplacian_line_detector.launch
ebretl: None of the line or cone detector launch files should launch the state publisher. That is the job of the gazebo or platform launch file, or of a testing launch file which includes sedani_state_publisher.launch for use with bag files. The same goes for the pointcloud converter launch file as well in the line below this one.

rr_iarrc/launch/cone_detection_height.launch
```diff
@@ -0,0 +1,28 @@
+<launch>
+    <node name="cone_detection_height" pkg="rr_iarrc" type="cone_detection_height" output="screen">
+
+        <param name="orange_low_H" type="int" value="5" />
+        <param name="orange_high_H" type="int" value="20" />
+        <param name="orange_low_S" type="int" value="150" />
+        <param name="orange_low_V" type="int" value="140" />
+
+        <param name="canny_cut_min_threshold" type="int" value="60" />
```
btdubs: Is this canny's minimum threshold and if so can we get canny's maximum as well as a knob to turn? Leave a note that the max is usually 2-3 times the value. Also, the cut term here is confusing because it usually used for removing things.

rr_iarrc/launch/cone_detection_height.launch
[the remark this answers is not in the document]
Daniel-Martin576: Where is it being published? Are you referring to the 643.31 in the projection_matrix within the camera_calibration file?

btdubs:
As a note @Daniel-Martin576 , you can just push to a branch you have a PR for and it will update. Reopening this and closing the new PR since they are the same but this one has comments.

Daniel's note from the new PR:
"I fixed the files with the changes desired from the last PL. I also removed the cone_detection_height because it has a lot of stuff that needs to be changed and optimized. In addition, I changed the cone_detecton_bottom to just loop through the columns and find the first white pixel (Originally, I thought this was slower while looking at timings, but enabling optimizations on catkin_make made it go significantly faster)."

Daniel-Martin576:
My bad, I thought the newer commits would be added on the top of the pull request, so when they weren't appearing I was confused and thought something wasn't working. Thank you for closing the other one for me.

PR #238, opened 2019-04-04
Renamed cone detection and laplacian nodes and Documented code [FIXED]

Daniel-Martin576:
I fixed the files with the changes desired from the last PL. I also removed the cone_detection_height because it has a lot of stuff that needs to be changed and optimized. In addition, I changed the cone_detecton_bottom to just loop through the columns and find the first white pixel (Originally, I thought this was slower while looking at timings, but enabling optimizations on catkin_make made it go significantly faster).


btdubs:
Closing in favor of using the old PR that has reviews. See https://github.com/RoboJackets/roboracing-software/pull/236

Issue #221, opened 2019-02-25
IARRC pointcloud combination with timestamp correlation


ebretl:
We have a node called the "mapper" in rr_avc. It subscribes to multiple pointclouds, and each time it receives new input points, it creates a new combined pointcloud in the base_footprint coordinate frame.

With separate line and cone detectors running at different speeds for the IARRC obstacle avoidance event, we should extend this mapper to publish pointclouds where each output is from a single source image. We will need to store a queue of input pointclouds for each source topic. We also need to make sure all parts of the perception pipeline (all detectors, perspective transformer, and pointcloud converter) preserve the original timestamp of the camera image message. 

While we're at it, the mapper should be moved to rr_common since it's not just for AVC anymore

PR #239, opened 2019-04-07
Synchronize pointcloud messages

dmulloy2:
Addresses #221

### DIFF
--- a/rr_common/src/mapper/mapper.cpp
+++ b/rr_common/src/mapper/mapper.cpp
@@ -13,12 +13,26 @@
 using cloud_t = pcl::PointCloud<pcl::PointXYZ>;
 using cloud_ptr_t = pcl::PointCloud<pcl::PointXYZ>::Ptr;
 
-std::map<std::string, sensor_msgs::PointCloud2ConstPtr> partials;
-bool needsUpdating = false;
+std::map<std::string, std::queue<sensor_msgs::PointCloud2ConstPtr>> queues;
+bool hasChanged = false;
+
+bool shouldUpdate() {
+    if (!hasChanged) {
+        return false;
+    }
+
+    for (const auto& entry : queues) {
+        if (entry.second.empty()) {
+            return false;
+        }
+    }
+
+    return true;
+}
 
 void cloudCallback(const sensor_msgs::PointCloud2ConstPtr &msg, std::string topic) {
-    partials[topic] = msg;
-    needsUpdating = true;
+    queues[topic].push(msg);
+    hasChanged = true;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


btdubs:
Can you also add an IARRC launch file for this node that will sub to the laplacian lines transformed cloud and the bottom cone transformed cloud topics?

Also, I know you didn't touch it, but can you remove the custom `split()` function and use `boost::split()` that uses `boost::is_any_of(" ,")` so that we can use space, comma or both. If we type weird, it will sometimes be size 0, so check for that. See https://github.com/RoboJackets/roboracing-software/blob/master/rr_common/src/image_pcl_converter/image_pcl_converter.cpp line 82.

rr_common/src/mapper/mapper.cpp
```diff
@@ -13,12 +13,26 @@
...
+std::map<std::string, std::queue<sensor_msgs::PointCloud2ConstPtr>> queues;
+bool hasChanged = false;
+
+bool shouldUpdate() {
+    if (!hasChanged) {
+        return false;
+    }
+
+    for (const auto& entry : queues) {
+        if (entry.second.empty()) {
```
btdubs: If I am not mistaken, this will only update everything if all of the topics had new data in the queue. It is possible that we are subscribed to a lot of topics but only a few are updated that should still be merged. Is that right? @ebretl

dmulloy2: That was my understanding of the new desired behavior, could we get clarification on that @ebretl ?

ebretl: We want to publish outputs that synchronize data from multiple inputs. As such, I think Dan's implementation is fine. To address the issue Brian raised, a good additional feature would be to require new data only on topics for which we have ever received a message, or for which we have received data in the past X seconds.

ebretl:
The code here works, but it doesn't fully solve the problem (that's my bad, probably, for not explaining the problem fully). So here goes:

Use case 1 (what I tried to explain originally): multiple streams of data (from line detector, cone detector) are being published based on the same camera data. The timestamps of each of these input pointclouds should be same as the camera image (I think I fixed this recently). So we want to combine the most recent set of pointclouds we have received which are all from the same camera frame. The code in this PR works only as long as an input pointcloud comes in from both processing streams for every camera frame. Usually this is the case, but it is not guaranteed. The mapper should throw out old input pointclouds.

Use case 2 (what I failed to anticipate earlier): we have data coming in from multiple cameras at the same time. Even when the frame rates of each camera are set to the same number, there is some variation in the actual rate of acquiring images. Also, the frame timestamps from each camera won't line up exactly. Say input one has data with timestamps T=10 and T=15, and input 2 has data T=7 and T=11. We should publish the combination of T=11 (the oldest queue head) and T=10, throw out T=7, and keep T=15 in the queue until more data on input 2 comes in.

rr_common/src/mapper/mapper.cpp
```diff
@@ -89,15 +103,15 @@ int main(int argc, char** argv) {
...
 
             combo_cloud->clear();
             combo_cloud_filtered->clear();
 
-            for(const auto& partial_pair : partials) {
-                // grab the reference to the sensor message from the map
-                auto& msg = partial_pair.second;
-                if(msg->width == 0) continue;
+            for(const auto& entry : queues) {
+                std::queue<sensor_msgs::PointCloud2ConstPtr> queue = entry.second;
```
ebretl: Ideally, we can avoid copying the whole queue like this

rr_common/src/mapper/mapper.cpp
```diff
@@ -89,15 +103,15 @@ int main(int argc, char** argv) {
...
+                auto& msg = queue.front();
+                queue.pop();
```
ebretl: msg is a reference to data which no longer guaranteed to exist (bc it has been popped from the container). Not sure how this works

btdubs:
I know we discussed this earlier and came up with a fix, but we may look at using this.
[http://wiki.ros.org/message_filters#ApproximateTime_Policy](http://wiki.ros.org/message_filters#ApproximateTime_Policy)

btdubs:
The issue was resolved as part of larger PR for IARRC competition 2019 in #247

PR #241, opened 2019-04-16
Gazebo iarrc upgrades

luisp23:
Makes gazebo modifications for IARRC.


rr_iarrc/src/urc_controller/urc_controller.cpp
```diff
@@ -0,0 +1,242 @@
...
+}
+
+// Used to calculate the distance from the target IMU yaw to the current IMU yaw within range of [0 ,6.28]
+bool comp(int a, int b) { 
+    return (a < b); 
+}
+
+float angleDist(float target, float current){
+    float diff = target - current;  
+    float min = std::min({std::abs(diff), std::abs(diff - 2*PI),std::abs(diff + 2*PI)}, comp);
```
ebretl: comp(a, b) takes in integers? Also, I think "a < b" is the default comparator, in which case min() shouldn't need to have it passed in

PR #244, opened 2019-05-18
Fixing cone height detector

Daniel-Martin576:
I cleaned up the cone_height_detector and added a publisher which publishes the distance to the closest cone.

### DIFF
--- a/rr_iarrc/launch/cone_bottom_detector.launch
+++ b/rr_iarrc/launch/cone_bottom_detector.launch
@@ -1,5 +1,5 @@
 <launch>
-    <node name="cone_bottom_detector" pkg="rr_iarrc" type="cone_bottom_detector" output="screen">
+<node name="cone_bottom_detector" pkg="rr_iarrc" type="cone_bottom_detector" output="screen">
         <!--HSV color threshold for orange cones-->
         <param name="orange_low_H" type="int" value="5" />
         <param name="orange_high_H" type="int" value="15" />
--- a/rr_iarrc/launch/cone_bottom_detector.launch
+++ b/rr_iarrc/launch/cone_bottom_detector.launch
@@ -13,4 +13,5 @@
 
         <param name="subscription_node" type="string" value="/camera/image_color_rect"/>
     </node>
+    <!-- In order to visualize resultant pointcloud, the line_pcl_converter node must be also be launched -->
 </launch>
--- a/rr_iarrc/src/cone_detection/cone_height_detector.cpp
+++ b/rr_iarrc/src/cone_detection/cone_height_detector.cpp
@@ -64,7 +64,9 @@ void publish_closest_point() {
                 min = cone_point.x;
             }
         }
-        pub_closest_point.publish((float) min);
+        std_msgs::Float32 msg;
+        msg.data = (float) min;
+        pub_closest_point.publish(msg);
     }
 }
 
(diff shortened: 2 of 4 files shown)


rr_iarrc/src/cone_detection/cone_height_detector.cpp
```diff
@@ -0,0 +1,379 @@
...
+pcl::PointCloud<pcl::PointXYZ> draw_cone_circles() {
+    pcl::PointCloud<pcl::PointXYZ> cone_cloud;
+    for (pcl::PointXYZ cone_point : cone_points) {
+        drawCircle(cone_cloud, cone_point, real_cone_radius, 20);
+    }
+    return cone_cloud;
+}
+
+
+void publishMessage(ros::Publisher pub, Mat img, std::string img_type) {
```
btdubs: Since you have a `publishPointCloud()` may want to call this `publishImage()`

At some point we.may want to create a standard RR library that has stuff like this so you don't have to copy paste every node.

rr_iarrc/src/cone_detection/cone_height_detector.cpp
```diff
@@ -0,0 +1,379 @@
...
+        threshold(dist_mask, thres, max * thres_value, 255, CV_THRESH_BINARY);
+        thres.convertTo(thres, CV_8UC1);
+        bitwise_or(thres, sure_front, sure_front);
+    }
+
+    return sure_front;
+}
+
+
+cv::Mat preform_watershed(const cv::Mat& frame, const cv::Mat& orange_found, const cv::Mat& detected_bodies) {
```
btdubs: Perform

rr_iarrc/launch/cone_bottom_detector.launch
```diff
@@ -1,5 +1,5 @@
 <launch>
-    <node name="cone_bottom_detector" pkg="rr_iarrc" type="cone_bottom_detector" output="screen">
+<node name="cone_bottom_detector" pkg="rr_iarrc" type="cone_bottom_detector" output="screen">
```
btdubs: Might want to note that this node is meant to be used along with the pcl overhead converter node.

rr_iarrc/src/cone_detection/cone_height_detector.cpp
```diff
@@ -0,0 +1,379 @@
...
+        ros::Duration(1).sleep();
+    }
+    ROS_INFO("Using horizontal FOV %f ", camera_fov_horizontal);
+}
+
+
+void publishPointCloud(pcl::PointCloud<pcl::PointXYZ> &cloud) {
+    sensor_msgs::PointCloud2 outmsg;
+    pcl::toROSMsg(cloud, outmsg);
+    outmsg.header.frame_id = "base_footprint";
```
btdubs: Is the point cloud output actually relative to the base_footprint? I think it is relative to the camera link.

rr_iarrc/src/cone_detection/cone_height_detector.cpp
```diff
@@ -0,0 +1,379 @@
...
+        cv_ptr->encoding = img_type;
+        cv_ptr->toImageMsg(outmsg);
+        pub.publish(outmsg);
+    }
+}
+
+
+cv::Mat draw_box_and_calc_dist(const cv::Mat& frame, const cv::Mat& sure_bodies) {
+    vector<vector<Point>> contours;
+    findContours(sure_bodies, contours, CV_RETR_TREE, CV_CHAIN_APPROX_SIMPLE);
```
btdubs: This should have the cv namespace attached. I think the compiler is just being nice and not complaining about it. There are lots of these type things with like cv::Point, watershed, etc. You may want to add them back to help with clarity but not a necessity.

Issue #242, opened 2019-04-24
navigation controller needs work


ebretl:
It has hard-coded values for AVC. Add launch file parameters, and optionally move to IARRC package

PR #245, opened 2019-05-31
Iarrc Navigation Controller and Finish Line Watcher

btdubs:
Tweaks the nav controller to not wait after the finish is detected. We had issue #242 but everything looks pretty general and usable for iarrc.

This also updates the finish line watcher using what Evan and I tested before school ended. The finish line watcher now uses the hsv tuner and color matcher instead of its own internal stuff.


btdubs:
After testing  yesterday, I am considering having the navigation controller allow steering for a small amount of time after finishing to ensure that the car stays within its lanes after crossing. We noticed that ending slightly angled after the end of a drag race will end with us stopping halfway outside of the lane. Otherwise, these changes worked during testing.

btdubs:
Merged as part of larger PR for IARRC competition 2019 in #247

PR #246, opened 2019-06-03
IARRC startlight node

btdubs:
The old stoplight watcher reacts more to people walking by than the stoplight itself. I reworked it to react only to circles which should require less tuning.

The new version using simple rgb math (no color matching) over the whole image to get the green and red light candidates. Once found, contours determine candidate green lights based on circularity. If a green light candidate is found, it looks in the past for a similarly sized red light above and nearby that position.

I also renamed the node to startlight_watcher to fix confusion since the light is made for starting races.


rr_iarrc/src/startlight_watcher/startlight_watcher.cpp
```diff
@@ -0,0 +1,224 @@
...
+        std::vector<cv::Vec3f> redCircles = findCirclesFromContours(redImage);
+        for (int i = 0; i < greenCircles.size(); i++) {
+            cv::Point greenCenter(static_cast<int>(greenCircles[i][0]), static_cast<int>(greenCircles[i][1]));
+            double greenRadius = greenCircles[i][2];
+
+            for (int j = 0; j < redCircles.size(); j++) {
+                cv::Point redCenter(static_cast<int>(redCircles[j][0]), static_cast<int>(redCircles[j][1]));
+                double redRadius = redCircles[j][2];
+                if (greenCenter.y > redCenter.y && //green below red
+                    std::fabs(greenRadius - redRadius) <= radiusTolerance &&
```
Daniel-Martin576: I think that x, y could be converted to col, row since greenCenter.row > redCenter.row means that green is below red (instead of greenCenter.y > redCenter.y) but it isn't completely necessary to change the code to incorporate this.

rr_iarrc/src/startlight_watcher/startlight_watcher.cpp
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    //Convert msg to Mat image
+    cv_bridge::CvImagePtr cv_ptr = cv_bridge::toCvCopy(msg, "bgr8");
+    cv::Mat frame = cv_ptr->image;
+
+    //gather a set of frames to run on
+    for(int i = history.size()-1; i > 0; i--) {
+        history[i] = history[i - 1];
+    }
```
Daniel-Martin576: Probably could just change this to use std::vector's insert() and pop_back() function if you wanted to simplify this.

Daniel-Martin576:
Good job! Worked well with the bag file I tested it on and I like that there is the ability to change the history buffer length. Minor comments but looks good enough to merge in my opinion.

rr_iarrc/src/startlight_watcher/startlight_watcher.cpp
btdubs: It definitely isn't intuitive at a glance, but (0,0) in the top left is the standard and the change away from cv::Point would probably clutter code. I think the comment is enough to make it clear, so I will leave this.

btdubs:
I don't think the variables redImage and redCircles can't be reused as they could not exist if there are no green circles found in the frame.
Hadn't heard of std::deque but that is exactly what I was looking for! Thank you. I wanted to leave it easy to implement checking multiple frames in the history instead of only the last in the buffer in the future which std::queue didn't allow. Changing to deque.

btdubs:
Merged as part of larger PR for IARRC competition 2019 in #247

PR #247, opened 2019-08-05
Competition2019

Daniel-Martin576:
Large merge of the cleaned 2019 competition code.


btdubs:
For future reference, here are major changes.
-Adds Sedani side cameras
-New start light detector: RGB math and circle detection
-New drag race controller: Finds 2 straight lines, PID to stay centered
-New Urban Race Challenge state machine controller
-New URC lane keeper utilizing drag race controller
-New URC sign detector and stop bar detector
-Updated point cloud combiner
-Adds obstacle history mapper for pointclouds

rr_common/src/image_pcl_converter/image_pcl_converter.cpp
```diff
@@ -64,7 +64,7 @@ void transformedImageCB(const sensor_msgs::ImageConstPtr& msg, const string& top
     sensor_msgs::PointCloud2 cloud_msg;
     pcl_conversions::fromPCL(cloud_pc2, cloud_msg);
     cloud_msg.header.frame_id = "base_footprint";
-    cloud_msg.header.stamp = ros::Time::now();
+    cloud_msg.header.stamp = msg->header.stamp;
```
ebretl: This is something we should do more of / continue to do going forward. If we are processing timestamped sensor data, the time we care about for the result is probably when the sensor made the measurement, not when the processing happened

rr_common/include/rr_common/RelativePoseHistoryClient.h
```diff
@@ -0,0 +1,35 @@
...
+   * @param t past time
+   * @return Pose (x, y, theta) in current local frame
+   */
+  Pose GetRelativePoseAtTime(const ros::Time& t);
+
+  /**
+   * Register callback for /pose_history topic
+   * @param handle NodeHandle to use for subscription
+   */
+  ros::Subscriber RegisterCallback(ros::NodeHandle& handle);
```
ebretl: Makes more sense to register the callback in the constructor, then it's impossible to construct a RelativePoseHistoryClient that isn't subscribed to a topic

rr_common/include/rr_common/RelativePoseHistoryClient.h
```diff
@@ -0,0 +1,35 @@
...
+#include <ros/ros.h>
+#include <geometry_msgs/Pose2D.h>
+#include <nav_msgs/Path.h>
+
+
+namespace rr {
+
+class RelativePoseHistoryClient {
+public:
+  using Pose = geometry_msgs::Pose2D;
```
ebretl: This type renaming is of no use. Pull the Pose2D name from geometry_msgs, and do this in the .cpp file

rr_platform/launch/sedani_cameras.launch
```diff
@@ -0,0 +1,29 @@
...
+    </node>
+
+    <include file="$(dirname)/camera_pipeline.launch">
+        <arg name="side" value="center"/>
+        <arg name="type" value="flir_blackfly"/>
+        <arg name="serial" value="18064461"/>
+        <arg name="manager_launched" value="true"/>
+    </include>
+
+    <!-- <include file="$(dirname)/camera_pipeline.launch">
```
ebretl: need better to way to specify cameras other than commenting out here

rr_common/include/rr_common/RelativePoseHistoryClient.h
ebretl: future improvement

rr_common/include/rr_common/RelativePoseHistoryClient.h
ebretl: future improvement

rr_iarrc/src/urc_controller/urc_controller.cpp
```diff
@@ -6,50 +6,56 @@
...
 double turn_speed;
-double left_turn_steering; 
-double right_turn_steering; 
+double left_turn_steering;
+double right_turn_steering;
 
 double forward_timeout;
 
-std::string turn_direction = "right";
+const std::string RIGHT("right");
```
ebretl: add these to constants to msg file so that sign detector agrees on state values

Daniel-Martin576:
- Delete/Archive rr_avc
- Add filtering node (either right when pointcloud is obtained or after summation)

PR #250, opened 2019-08-29
fix camera geometry not loading bug

ebretl:
Found what seems to be a bug in the pointcloud camera projector which caused it to hang while waiting for a thread to finish. Found while testing in Gazebo. Perhaps the issue was ros::spinOnce() not working within the other thread? In any case, we can accomplish waiting for that thread by returning from all image callbacks early. (This is what I get for using multiple threads I guess)


Daniel-Martin576:
Looks good

PR #252, opened 2019-09-07
Fixed README simulator launch file

Daniel-Martin576:
I simply changed the navigation planner used on our README because the old ```avc_sim.launch``` wasn't working due to recent file changes.


ebretl:
The directions still say to pull up the AVC track. Is that what we want? If not, we have a working configuration that gets around the IARRC sim track

Daniel-Martin576:
That was intentional, but I am open to the idea of having it run on the IARRC sim track so we can start to potentially separate ourself from AVC files since we aren't doing the competition anymore.

Daniel-Martin576:
I agree

Issue #253, opened 2019-09-07
Delete rr_avc directory


Daniel-Martin576:
Since AVC is no longer one of our competitions, the files within the ```rr_avc``` directory must be moved or deleted. Simply move useful files to existing related folders or completely deleted them if not.

PR #261, opened 2019-09-19
Deletes rr_avc, creates rr_msgs, and cleans up the resulting mess

### LINKED ISSUE #258: restructure avc and msgs packages
- delete rr_avc
- consolidate messages into rr_msgs folder
- fix package dependency tree in general (catkin_make_isolated should succeed, but not if we remove anything)

ebretl:
Fixes #258 #253 

- deletes rr_avc
- creates rr_msgs
- I was partially wrong about rr_msgs making our CMakeLists.txt files cleaner. When building with `catkin build` or `catkin_make_isolated`, we no longer need to `add_dependencies` from all our CMake targets that require our custom messages. However, `catkin_make` requires still placing these `add_dependencies` calls all throughout the repo :upside_down_face: 
- Also took the time with our package.xml files to convert them to format "2" (which has the handy <depend> tag combining <build_depend> and <exec_depend>) and to remove unnecessary build dependencies

Somehow that managed to touch almost every file in the repo


rr_common/src/image_transformation/pointcloud_projector.cpp
```diff
@@ -11,7 +11,7 @@
 #include <sensor_msgs/Image.h>
 #include <sensor_msgs/PointCloud2.h>
 
-#include <rr_platform/CameraGeometry.h>
+#include "../../include/rr_common/CameraGeometry.h"
```
Daniel-Martin576: Would it be possible to do a global path instead of a local one here?

Daniel-Martin576:
1. Merge master and use ```make format```
2. ```race_reset.msg``` file is empty... Is there supposed to be something there?
Looks good though!

PR #265, opened 2019-09-23
add ground-truth gz plugin and robot_localization minimal config

### LINKED ISSUE #263: EVGP simulate perfect localization
Use a ground-truth plugin or similar to get the perfect relative pose of base_footprint with respect to our starting point. Use for mapping experiments, verifying that state estimating converges, etc.

ebretl:
Fixes #263

### DIFF
--- a/rr_description/urdf/evgp.urdf.xacro
+++ b/rr_description/urdf/evgp.urdf.xacro
@@ -300,6 +300,20 @@
         </plugin>
     </gazebo>
 
+    <!-- ground-truth localization plugin -->
+    <gazebo>
+        <plugin name="ground_truth_tracker" filename="libgazebo_ros_p3d.so">
+            <alwaysOn>true</alwaysOn>
+            <updateRate>50.0</updateRate>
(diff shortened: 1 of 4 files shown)


Daniel-Martin576:
Looks good, the first time I ran the launch file I was getting errors concerning TF_OLD_DATA but it worked when I reset all the nodes.

Daniel-Martin576:
Some notes to self:
* I kinda wanna reduce the number of topics being published to which aren't being used
* /camera/image_color_rect/compressed, theora, compressedDepth is still being published even with         disable_pub_plugins in launch file
* The /gps has 8 topics related to it, not sure if all are necessary

ebretl:
- Hmm I thought the disable_pub_plugins thing was working
- yeah, when using the world ref frame I always disable either the RobotModel or Camera plugins. Not sure of any way around that
- I've tried my best to balance weird bounciness/instability issues with the gazebo server's CPU usage. The way around that might be making a simplified/low-polygon collision mesh. I'll leave that to another issue

PR #266, opened 2019-09-23
Basic mapping using costmap_2d

### LINKED ISSUE #262: EVGP mapping in simulation - basic
Set up something to generate/update a map assuming perfect localization and only static obstacles. This are unrealistic assumptions, of course, but the goal is to get something ready to test downstream functionality (e.g. map post-processing and planning)

ebretl:
Fixes #262 
Merge #265  first (this branch builds on that one)


Daniel-Martin576:
Great job, I really like the video that came out of this. 
For some reason Detection in rviz for /current_obstacle sometimes gives a transform error of 
```For frame [base_footprint]: No transform to fixed frame [world].  TF error: [Lookup would require extrapolation into the future.  Requested time 752.943000000 but the latest data is at time 752.938000000, when looking up transform from frame [base_footprint] to frame [world]]```
It doesn't seem to be a big problem though

Daniel-Martin576:
You have varying amounts of frequencies throughout the mapping and localization code.  We don't need to change them now since it is just simulation and experimentation, but maybe when we get the car we should test which rate works better (faster with good results).

PR #269, opened 2019-10-04
Simulating dynamic obstacles

### LINKED ISSUE #264: EVGP simulate dynamic obstacles
Create "opponent" cars which follow either a pre-set trajectory of points or follow a simple control law. They must be detectable by lidar/camera, but ideally collisions should be optional

ebretl:
Fixes #264 

- Adds EVGP car opponent URDF. The URDF file overlap between the real car and the opponent is moved to a common file evgp_body.urdf.xacro
- Adds a model controller ROS node which reads a trajectory from a CSV file and applies positions and velocities via Gazebo's SetModelState service
- Adds a script which reads a rosbag and produces a suitable CSV file for the model controller
- Removes obsolete code from rr_platform/CMakeLists.txt


Daniel-Martin576:
Looks good! Really like how it turned out and the feature to convert bag files to trajectory CSVs.
1. Please change the opponent bots to be non-collidable. Preferably have this controlled by a simple boolean in its xacro file. 
2. Move the opponent car position back or just move our starting position forward so that we navigate through the course with the other car next to us. Not completely necessary, but would help with testing.

Daniel-Martin576:
The opponent is non-collidable now but it doesn't seem to be picked up by the mapping.
Also, we should consider removing the gpu  from gpu_ray and gpu_laser in evgp.urdf.xacro for those without a gpu.

oswinso:
I think a parameter for `gpu` default false will probably be better

Daniel-Martin576:
Actually don't merge yet, mapping with sensor is behaving kinda strangely

PR #274, opened 2019-10-11
Adds bayes filter obstacle layers

ebretl:
This PR adds the bayes filter mapper as another option if we want it. The default in mapping_from_ground_truth.launch is still the "plain old obstacle plugin"


rr_evgp/src/binary_bayes_filter_obstacle_layer.cpp
```diff
@@ -0,0 +1,250 @@
...
+            double march_x = lidar_x_;
+            double march_y = lidar_y_;
+            const double last_clear = range - resolution;
+            while (d < last_clear) {
+                master_grid.worldToMapNoBounds(march_x, march_y, mx, my);
+                updates_[master_grid.getIndex(mx, my)] = 0;
+                d += step_size;
+                march_x += step_x;
+                march_y += step_y;
+            }
```
oswinso: Interesting that you do it this way. How is the performance like?

rr_evgp/src/binary_bayes_filter_obstacle_layer.cpp
```diff
@@ -0,0 +1,250 @@
...
+                if (update >= 0) {
+                    double p_prior = probs_[i];
+                    if (update > 0) {  // hit
+                        double p_occupied = (1.0 - prob_false_neg_) * p_prior;
+                        double p_empty = prob_false_pos_ * (1.0 - p_prior);
+                        probs_[i] = std::min(p_occupied / (p_occupied + p_empty), max_prob_);
+                    } else {  // cleared
+                        double p_occupied = prob_false_neg_ * p_prior;
+                        double p_empty = (1.0 - prob_false_pos_) * (1.0 - p_prior);
+                        probs_[i] = std::max(p_occupied / (p_occupied + p_empty), 1.0 - max_prob_);
```
oswinso: no log odds notation? I think it'll also be more numerically stable

rr_evgp/src/binary_bayes_filter_obstacle_layer.cpp
```diff
@@ -0,0 +1,250 @@
...
+                    }
+                }
+
+                costmap_data[i] = static_cast<uint8_t>(probs_[i] * 255.999);
```
oswinso: What's the rationale for 255.999 instead of 255?

ebretl: Rounding down so only 1.0 would be mapped to 255, but the max_prob limit doesn't let it get to 1.0

rr_evgp/src/binary_bayes_filter_obstacle_layer.cpp
ebretl: Log odds would be a good future improvement

rr_evgp/src/binary_bayes_filter_obstacle_layer.cpp
ebretl: I haven't profiled this method vs costmap_2d's Bresenham implementation, but overall this layer has the same or less CPU usage than ObstacleLayer. My raytracing loop executes more times but has no branches

PR #280, opened 2019-10-29
Bayes filter mapping

ebretl:
We approved and merged this branch into the 264-dynamic-obst-sim branch instead of master ( #274 ). I reverted this change on that branch and am re-opening this PR into master. The changeset in this PR is a superset of #269, so that one should be merged first.


rr_evgp/src/binary_bayes_filter_obstacle_layer.cpp
```diff
@@ -0,0 +1,250 @@
...
+            for (int my = min_cell_y; my < max_cell_y; my++) {
+                updates_[master_grid.getIndex(mx, my)] = -1;
+            }
+        }
+
+        double angle = scan_->angle_min;
+        for (double raw_range : scan_->ranges) {
+            bool is_hit = (!std::isinf(raw_range) && raw_range <= scan_range_);
+            double range = is_hit ? raw_range : scan_range_;
+            double trace_range = range + (resolution * 0.7);
```
Daniel-Martin576: Why did you choose to multiply resolution by .7?

ebretl: When the localization drifts around, treating an obstacle as something with a bit of depth (so marking more than 1 cell for a single lidar point) generates nicer-looking maps. I just pushed an update to make it more clear in the code what is going on with this.

PR #283, opened 2019-11-12
Parallel hill climbing for IARRC planner

### LINKED ISSUE #248: Nearest point to possibly inscribed set is added on every iteration
https://github.com/RoboJackets/roboracing-software/blob/fcc0968fc6f41deb73153dee1a39b999f47a281f/rr_common/src/planner/distance_checker.cpp#L163-L165

The above lines I think are unecesary since
https://github.com/RoboJackets/roboracing-software/blob/fcc0968fc6f41deb73153dee1a39b999f47a281f/rr_common/src/planner/distance_checker.cpp#L111-L120

already adds all `might_hit_points` of the parent to the child during `SetMap`, and the `nearest_point` is added to the "root" cell's `might_hit_points` here: 
https://github.com/RoboJackets/roboracing-software/blob/fcc0968fc6f41deb73153dee1a39b999f47a281f/rr_common/src/planner/distance_checker.cpp#L79

This also results in the closest node being added to `might_hit_points` every time `GetCollisionDistance` is called, resulting in duplicate instances of the closest node, as noticed by @ebretl.

ebretl:
It seems like this method creates paths which are similar to or better than simulated annealing with about half the latency. :tada: 

Notes:
- still using pointcloud input
- spawns M new threads every iteration. This could be optimized but I don't think that would be worth it.
- if more restarts than threads are desired, each thread assigns itself more tasks based on availability
- moves responsibility of retrieving planner params (the ones that may or may not be used, depending on `planner_type`) from `planner_node` to each planner class

Side note: In the process of doing this, I think I figured out that issue #248 was not a major performance issue as I thought but more of a readability issue. The nearest point was being repeatedly added and removed from the list of possible collision points. Making `GetCollisionDistance(pose)` be `const` now keeps that from happening. Fixes #248


rr_common/src/planner/annealing_planner.cpp
```diff
@@ -85,28 +97,24 @@ PlannedPath AnnealingPlanner::Plan(const PCLMap& map) {
...
-            has_been_too_close |= (collision || dist < 0.5);
-            if (has_been_too_close) {
-                path.path[i].speed *= 0.7;
-            }
+        // Check collisions
+        path.has_collision = false;
+        for (size_t i = 0; i < path.path.size(); ++i) {
+            auto opt_dist = distance_checker_.GetCollisionDistance(path.path[i].pose);
+            path.has_collision |= !opt_dist.has_value();
+            path.dists[i] = opt_dist ? *opt_dist : -1;
```
oswinso: I think `opt_dist.value_or(-1)` would be more clear

rr_common/src/planner/hill_climb_planner.cpp
```diff
@@ -0,0 +1,148 @@
...
+PlannedPath HillClimbPlanner::Plan(const rr::PCLMap& map) {
+    distance_checker_.SetMap(map);
+
+    auto descend_hill = [this, &map](PlannedPath& plan) {
+        plan.has_collision = false;
+        // if control is not inited externally, select a random starting configuration
+        if (plan.control.size() != segment_sections_.size()) {
+            plan.control.resize(segment_sections_.size());
+            std::fill(plan.control.begin(), plan.control.end(), 0.0);
+            JitterControls(plan.control, max_steering_ / 2);
```
Daniel-Martin576: Why did you set `stddev` to `max_steering_ / 2`?

rr_common/src/planner/distance_checker.cpp
```diff
@@ -148,65 +148,57 @@ std::tuple<bool, double> DistanceChecker::GetCollisionDistance(const Pose& pose)
...
-            entry.might_hit_points.push_back(entry.nearest_point);
+    const CacheEntry& entry = cache_[i];
+
+    auto point_in_local_frame = [search_x, search_y, cos_th, sin_th](const PCLPoint& p) {
+        double offsetX = p.x - search_x;
+        double offsetY = p.y - search_y;
+        double x = cos_th * offsetX + sin_th * offsetY;
+        double y = -sin_th * offsetX + cos_th * offsetY;
+        return std::make_tuple(x, y);
+    };
```
oswinso: Is there a reason why you don't use a `tf::Transform` and transform the point that way?

rr_common/src/planner/hill_climb_planner.cpp
```diff
@@ -0,0 +1,148 @@
...
+                stuck_counter = local_optimum_tries_;
+                best_cost = plan.cost;
+            }
+        }
+
+        // ROS_INFO_STREAM("descent took " << jitter_counter << " rollouts");
+    };
+
+    std::vector<PlannedPath> best_plans(num_workers_);
+    for (PlannedPath& plan : best_plans) {
```
Daniel-Martin576: Any reason you don't just keep the best path and change it when you have the lock?

rr_common/src/planner/hill_climb_planner.cpp
```diff
@@ -0,0 +1,148 @@
...
+        } else if (x > max_steering_) {
+            x = max_steering_;
+        }
+    }
+}
+
+PlannedPath HillClimbPlanner::Plan(const rr::PCLMap& map) {
+    distance_checker_.SetMap(map);
+
+    auto descend_hill = [this, &map](PlannedPath& plan) {
```
Daniel-Martin576: Can't you just do `[&]`?

rr_common/src/planner/hill_climb_planner.cpp
```diff
@@ -0,0 +1,148 @@
...
+        plan.cost = std::numeric_limits<double>::max();
+    }
+    int plan_count = 0;
+    std::mutex plan_count_mutex;
+
+    auto worker = [&, this](int thread_idx) {
+        while (true) {
+            {
+                std::lock_guard lock(plan_count_mutex);
+                if (plan_count >= num_restarts_) {
```
Daniel-Martin576: You could do this comparison before/without getting the lock since it is only a read

rr_common/include/planner/annealing_planner.h
```diff
@@ -3,6 +3,8 @@
 #include <random>
 #include <vector>
 
+#include <ros/node_handle.h>
```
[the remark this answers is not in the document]
ebretl: The idea is that this file only uses the NodeHandle class, not all of ROS. Unfortunately node_handle.h includes most of ROS anyway

rr_common/include/planner/hill_climb_planner.h
```diff
@@ -0,0 +1,46 @@
...
+    double neighbor_stddev_;
+    int local_optimum_tries_;
+
+    DistanceChecker distance_checker_;
+    BicycleModel model_;
+
+    std::normal_distribution<double> normal_distribution_;
+    std::mt19937 rand_gen_;
+
+    PlannedPath previous_best_path;
```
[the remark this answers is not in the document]
ebretl: Yes, I'll add that for consistency

rr_common/src/planner/distance_checker.cpp
ebretl: It's this way because I already had this code, and just refactored it

rr_common/include/planner/distance_checker.h
```diff
@@ -29,7 +30,7 @@ class DistanceChecker {
      * @return bool - collision detected
      *         double - clearing distance to nearest obstacle
      */
-    std::tuple<bool, double> GetCollisionDistance(const Pose& pose);
+    [[nodiscard]] std::optional<double> GetCollisionDistance(const Pose& pose) const;
```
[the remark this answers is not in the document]
ebretl: I put nodiscard to make CLion clang-tidy chill out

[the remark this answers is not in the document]
ebretl: I was hoping the optional would be a clearer interface, but in hindsight that might not be true

rr_common/src/planner/hill_climb_planner.cpp
```diff
@@ -0,0 +1,148 @@
...
+}
+
+void HillClimbPlanner::JitterControls(std::vector<double>& ctrl, double stddev) {
+    for (double& x : ctrl) {
+        x += normal_distribution_(rand_gen_) * stddev;
+        if (x < -max_steering_) {
+            x = -max_steering_;
+        } else if (x > max_steering_) {
+            x = max_steering_;
+        }
```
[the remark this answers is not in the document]
ebretl: Will fix

rr_common/src/planner/hill_climb_planner.cpp
ebretl: This way was marginally easier for me to write. We can test that option too and see if it's any better or worse

rr_common/src/planner/hill_climb_planner.cpp
ebretl: I think there's a potential race condition if you check before the lock

rr_common/include/planner/annealing_planner.h
```diff
@@ -13,10 +15,10 @@ namespace rr {
 class AnnealingPlanner : public Planner {
   public:
     struct Params {
-        unsigned int n_path_segments;  // number of segments in a path
-        unsigned int annealing_steps;  // number of timesteps to run simulated annealing
-        double temperature_start;      // temperature parameter starts high and decreases
-        double temperature_end;
+        int n_path_segments;       // number of segments in a path
```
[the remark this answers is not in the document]
ebretl: Uh kind of because this is a struct or record type so its members won't really be confused with local variables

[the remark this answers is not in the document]
oswinso: Ah ok mb I thought these were member variables of `AnnealingPlanner` and not `Params`

rr_common/src/planner/hill_climb_planner.cpp
ebretl: I think we have to capture `this` by value

rr_common/src/planner/hill_climb_planner.cpp
ebretl: I tested it and your way seems to be 0.04ms faster on average over a 10 minute trial. I thought it was going to be slower

Daniel-Martin576: Do you think it would be faster if every worker kept there best path internally and didn't check the global one until it has finished all it can do? I don't think the change at this point would be very major, but it would have the lock for less time.

rr_evgp/conf/planner_sim_hillclimb.yaml
```diff
@@ -0,0 +1,30 @@
...
+    side: 0.65
+
+max_speed: 5.0
+max_drive_accel: 10
+
+wheel_base: 0.97
+lateral_accel: 2.0
+max_steering: 0.3
+steering_speed: 1000
+steering_gain: 1.0
```
Daniel-Martin576: Should this be added to every planner for IARRC as well? 
Because `roslaunch rr_iarrc test_circuit_sim.launch` with `roslaunch rr_gazebo macaroni_iarrc_circuit.launch` make the car drive off the course on the first turn now.

Daniel-Martin576: Lol, the circuit planner on master also seems to be broken...

Daniel-Martin576: Nvm, all is fixed when I update the laplacian line sim detector

PR #287, opened 2019-11-13
Local and Global Map with Gap Closure

### LINKED ISSUE #271: Map Closer
Add a layer to the cost map with additional barriers added to the holes along the wall of the track. A closed-circuit should be produced which will prevent our local planner from considering departing the track. 

Possible Implementation:
* Convert the map to an image (may need to do regardless for OpenCV distance transform if we decide to use just that for finding the centerline)
* Find connected contours 
* Find endpoints of contours
* Find line extension of contour at the endpoint (some sort of linear regression)
* Connect contours that have line extensions with certain interception angles or are close enough together 
* Maybe this would be helpful: https://stackoverflow.com/questions/35847990/detect-holes-ends-and-beginnings-of-a-line-using-opencv

This line extension process should be occurring live so the planner can use this information during the cold lap as well.

Daniel-Martin576:
Contains global and local costmaps along with the gap closer. Map closer speed is strongly related to map local size and the amount of dilatation. When testing the speed of the gap closure was around ~5 ms where skeletonization took about ~3-4 ms of that. The gap closure should only be run during the cold lap so there should be too many issues with speed since that lap will probably be slower. 
Fixes #271


rr_evgp/src/costmap_plugins/track_closing_layer.cpp
```diff
@@ -0,0 +1,248 @@
...
+
+    return x;
+}
+
+bool TrackClosingLayer::IsConnectable(const WallEndpointContext& current, const WallEndpointContext& prev) {
+    const auto& [endpntA, pntA, vxA, vyA] = current;
+    const auto& [endpntB, pntB, vxB, vyB] = prev;
+
+    double angle_between_lines = std::acos(vxA * vxB + vyA * vyB) * 180.0 / M_PI;
+    angle_between_lines = std::min(angle_between_lines, 180 - angle_between_lines);
```
ebretl: I kinda doubt that `min(angle, 180 - angle)` is what we want here. That implies that angle differences of 0 and 180 are equally good. We're looking for angles near 180 only, right?

rr_evgp/src/costmap_plugins/track_closing_layer.cpp
```diff
@@ -0,0 +1,248 @@
...
+        if (!have_connection[i]) {
+            const auto& endpnt = std::get<0>(curr_line_context);
+            const auto& extrapolated_pnt = std::get<1>(curr_line_context);
+            // cv::line(debug_img, endpnt, extrapolated_pnt, cv::Scalar(0, 0, 255));
+            cv::line(walls, endpnt, extrapolated_pnt, cv::Scalar(255));
+        }
+    }
+
+    // Make map from image
+    cv::morphologyEx(walls, walls, cv::MORPH_DILATE, dilate_kernel);
```
ebretl: I think the dilation here should use a smaller kernel than what is used before skeletonization. Or at least we should be able to change their sizes independently

rr_evgp/src/costmap_plugins/track_closing_layer.cpp
```diff
@@ -0,0 +1,248 @@
...
+
+void TrackClosingLayer::updateCosts(costmap_2d::Costmap2D& master_grid, int min_cell_x, int min_cell_y, int max_cell_x,
+                                    int max_cell_y) {
+    if (!enabled_) {
+        return;
+    }
+
+    // populate OpenCV Mat as grid
+    cv::Mat mat_grid(master_grid.getSizeInCellsX(), master_grid.getSizeInCellsY(), CV_8UC1);
+    uint8_t* charMap = master_grid.getCharMap();
```
ebretl: It's possible that the obstacle layer may edit masterGrid's underlying charMap asynchronously (something like this happens in ROS Kinetic). If so, we will need to add synchronization. So far I haven't seen the weird effects that I ran up against this summer, so that's good

ebretl:
@Daniel-Martin576 I'm going to go ahead and fix the things I identified in my review

rr_evgp/conf/mapping_local.yaml
```diff
@@ -0,0 +1,56 @@
...
+            inf_is_valid: false
+            obstacle_range: 7.99
+            raytrace_range: 7.98
+
+    inflation:
+        inflation_radius: 0.3
+        cost_scaling_factor: 1.0
+        lethal_cost_threshold: 1
+
+    distance_field:
```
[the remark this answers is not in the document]
ebretl: Resolving because this is accomplished in `feat/285-costmap-planning`

rr_evgp/src/costmap_plugins/track_closing_layer.cpp
ebretl: Going to leave this comment in the review, but won't fix it right now because it is not apparent what (if anything) needs to change

rr_evgp/CMakeLists.txt
```diff
@@ -6,29 +6,27 @@ set(CMAKE_CXX_STANDARD 17)
...
 
 
 include_directories(
 # include
   ${catkin_INCLUDE_DIRS}
+  ${OpenCV_INCLUDE_DIRS}
 )
 
-add_library(rr_evgp_plugins src/binary_bayes_filter_obstacle_layer.cpp)
-target_link_libraries(rr_evgp_plugins ${catkin_LIBRARIES})
```
Daniel-Martin576: Why were these removed?

ebretl: Pretty sure I moved them to another cmakelists file inside src

PR #288, opened 2019-11-19
Improved finish line watcher by analyzing contours

### LINKED ISSUE #277: Improve Finish Line Watcher
**Goal:** Edit current finish line watcher to more accurately detect the finish line. 

**Notes:** Since the event's finish line is usually a distinctive color, the current code simply use the HSV tuner to threshold for that particular color and then finds the total number of pixels and longest consecutive number of pixels in a row of that color. This caused issues during the competition because sometimes the sun or shadows would cause the colored finished line to appear split meaning it won't be detected. 
- A potential solution is simply finding the contours of the thresholded image and if the width of a contour is over a threshold then it is the finish line. 
- Another potential method is simply counting the number of columns of the whole image that have the colored pixel and incorporating some numeric threshold for the minimum number of columns needed to be found for the line to be considered a finish line.

Also, add a boolean in the launch file for publishing that a finish line has been detected either:
- Right when the finish is detected (useful in the drag race when we want to start breaking as early as possible) 
- Right after the car passes the finish line (useful in the circuit race when we want to make sure we stop after crossing the finish line). 

For testing the behavior of the finish line detector, use a bag file from `https://cloud.robojackets.org/` under /RoboRacing/Bag Files/2019/2019-iarrc-comp for either the circuit or drag race. 

Also, change the launch file name from `race_end_detector.launch` to `finish_line_detector.launch`

**Files Edited (at least):** `/rr_iarrc/launch/perception/race_end_detector.launch`, `/rr_iarrc/src/finish_line_watch/finish_line_watcher.cpp`

**Returning Member Mentor:** Daniel Martin @Daniel-Martin576  --Ask him questions!
When you believe you have finished the success criteria below, submit a pull request to merge your code into master.

**Success Criteria:**
- Able to run finish_line_detector
- HSV tuner node still dynamically adjusts what color is being thresholded for the finish line
- Finish line detected reliably in the previous competition's circuit and drag race bag files

dsiegler2000:
Fix #277 
Improved the finish line watcher. Still using the HSV thresholding layer. Instead of counting the width of the thresholded pixels, instead I perform contour detection. I then filter out very small contours. After that I find a rotated rectangle that fits all the remaining contours. I use the width and angle of that rectangle to determine if we have detected a finish line (must be relatively flat and wide).

Also, changed the name of the launch file to `race_end_detector.launch` and added a boolean to that file, `publish_when_detected`. If set to true, the finish lines crossed counter topic will tick up when the finish line is first detected. If set to false, the finish lines crossed counter topic will tick up when the finish line is crossed (i.e. when it goes from being detected to un-detected). 

This code was tested on iarrc19_2019-drag-end.bag and Sedani_2019-circuit-end.bag and worked on both. Video here: https://drive.google.com/file/d/1jOXLI-K7vioFdn95BInQYZouKdUaB4O6/view?usp=sharing

edited on 12/2 to reflect the changes in code.


rr_iarrc/src/finish_line_watcher/finish_line_watcher.cpp
```diff
@@ -78,23 +74,85 @@ void ImageCB(const sensor_msgs::ImageConstPtr& msg) {
...
-    auto width = getWidth(frame);
+    bool detected = maxLength > length_cutoff;
 
-    if (state == LOW && count > 2000 && width > 700) {
-        state = HIGH;
-    } else if (state == HIGH && count < 2000) {
-        state = LOW;
-        number_of_crosses++;
-        ROS_INFO_STREAM("Finish line crossed: " << to_string(number_of_crosses));
+    if (publish_when_detected) {
```
btdubs: Here, we want the functionality to always uptick the number of crosses. The difference is when it ups the value, either as soon as it sees a valid finish line or as soon as we lose sight of a seen finish line.

rr_iarrc/src/finish_line_watcher/finish_line_watcher.cpp
```diff
@@ -78,23 +74,85 @@ void ImageCB(const sensor_msgs::ImageConstPtr& msg) {
...
+        double slope = std::abs(rise / run);
+        // Also find the length
+        double length = arcLength(contours.at(i), false);
+
+        // Filter by slope and contour
+        if (slope > 0.0001 && slope < slope_cutoff && length > length_cutoff) {
+            // Draw for debug image
+            drawContours(drawing, contours, (int)i, Scalar(255, 0, 0), 2, LINE_8, hierarchy, 0);
+
+            if (length > maxLength) {
```
btdubs: Besides length, other shape factors like size could be important possibly. Note the old detector cared about the number of pixels.

rr_iarrc/src/finish_line_watcher/finish_line_watcher.cpp
dsiegler2000: I (think) that's what this code is doing; if we are upticking when the line is first detected (`publish_when_detected=true`) then state will be low and, if the cooldown is at 0 and the line is detected, `if (detected && state == LOW && cooldown == 0)` will be true so the value will be increased and then the state is reset by the following else if. If we are upticking when we lose sight of the line then the opposite happens: the state to high when the line is detected then when we lose sight of the line then we reset the state to low and increase the value.

rr_iarrc/src/finish_line_watcher/finish_line_watcher.cpp
```diff
@@ -78,23 +74,85 @@ void ImageCB(const sensor_msgs::ImageConstPtr& msg) {
...
+    Mat contourImg = frame.clone();
+    findContours(contourImg, contours, hierarchy, CV_RETR_LIST, CV_CHAIN_APPROX_SIMPLE);
+
+    Mat drawing = Mat::zeros(contourImg.size(), CV_8UC3);
+    double maxLength = 0.0;
+    double maxLengthYPos = 0.0;
+    for (size_t i = 0; i < contours.size(); i++) {
+        // Find the average slopes of the contour
+        Point start = contours.at(i).at(0);
+        Point end = contours.at(i).at((contours.at(i).size() - 1) / 2);
```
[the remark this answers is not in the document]
dsiegler2000: My intention of how this would work is the following:

- Look at the HSV thresholded image for the bright pink
- Detect _lines_ in the image
- Filter the lines as to get rid of those that are quite short or those that have a high slope. Looking at the bag files it looks like when looking for the finish line the car is coming in at more or less straight on, so we don't want to include any lines that are particularly vertical, thus adding the slope restriction.
- If there are still lines left after this filtering then I assume that's the finish line.

I now realize that I was looking at contours, not lines, which is why calculating slope seems fairly nonsensical. I will update to use line detection, not contour detection. It seemed that sometimes contour detection would split the finish line up into two contours, which is why I didn't want to use it.

rr_iarrc/src/finish_line_watcher/finish_line_watcher.cpp
```diff
@@ -78,23 +87,102 @@ void ImageCB(const sensor_msgs::ImageConstPtr& msg) {
     frame = cv_ptr->image;
     blockEnvironment(frame);
 
-    auto count = countNonZero(frame);
+    Mat src = frame.clone();
+    Mat dst;
+    cv::Canny(src, dst, canny_thresh1, canny_thresh2, 3);
```
btdubs: It will barely affect performance, but Canny is overkill when you already have a binary image. A simple laplacian is a better way to get the lines.

rr_iarrc/src/finish_line_watcher/finish_line_watcher.cpp
```diff
@@ -78,23 +84,113 @@ void ImageCB(const sensor_msgs::ImageConstPtr& msg) {
...
+        // When the line is first detected, set to a high state then when we stop detecting (we've crossed), report that
+        if (state == LOW && detected && cooldown == 0) {
+            state = HIGH;
 
-    auto width = getWidth(frame);
+        } else if (state == HIGH && !detected) {
+            state = LOW;
+            number_of_crosses++;
+            cooldown = cooldown_value;
+            ROS_INFO_STREAM("Finish line crossed: " << to_string(number_of_crosses));
```
Daniel-Martin576: Maybe compress this with the previous block so that you set `incrementCrossNum = True && publish_when_detected` in one block and `incrementCrossNum = True && !publish_when_detected` in the other.
Then have an if-statement later that checks incrementCrossNum where you actually set the number of crosses, the cooldown, and prints it. 
Just so it is easier to read.

rr_iarrc/src/finish_line_watcher/finish_line_watcher.cpp
```diff
@@ -78,23 +84,113 @@ void ImageCB(const sensor_msgs::ImageConstPtr& msg) {
...
 
-    auto count = countNonZero(frame);
+    Mat dst;
+    Mat drawing = Mat::zeros(frame.size(), CV_8UC3);
+    cv::cvtColor(frame, drawing, cv::COLOR_GRAY2BGR);
+
+    // Find contours
+    vector<vector<Point> > contours;
+    vector<Vec4i> hierarchy;
+    cv::Canny(frame, dst, canny_thresh1, canny_thresh2, 3);
```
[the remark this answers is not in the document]
dsiegler2000: The finish line watcher uses an HSV thresholding node so the input (`frame`) is already a binary image with the line thresholded.

rr_iarrc/src/finish_line_watcher/finish_line_watcher.cpp
```diff
@@ -78,23 +84,113 @@ void ImageCB(const sensor_msgs::ImageConstPtr& msg) {
...
+    cv::findContours(dst, contours, hierarchy, RETR_TREE, CHAIN_APPROX_SIMPLE);
+
+    // Filter out very small contours
+    vector<Point> filteredContourPoints;
+    for (int i = 0; i < contours.size(); i++) {
+        cv::Scalar color(0, 0, 255);
+
+        if (cv::contourArea(contours.at(i)) > min_contour_area) {
+            filteredContourPoints.insert(std::end(filteredContourPoints), std::begin(contours.at(i)),
```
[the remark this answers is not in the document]
dsiegler2000: It can't find a method signature for `push_back(std::vector)`

rr_iarrc/src/finish_line_watcher/finish_line_watcher.cpp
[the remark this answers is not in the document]
Daniel-Martin576: True, then why do you do canny instead of just findContours then?

rr_iarrc/src/finish_line_watcher/finish_line_watcher.cpp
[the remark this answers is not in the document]
Daniel-Martin576: True, forget you were dealing with vectors for a sec

rr_iarrc/src/finish_line_watcher/finish_line_watcher.cpp
```diff
@@ -78,23 +84,122 @@ void ImageCB(const sensor_msgs::ImageConstPtr& msg) {
...
+        cooldown--;
+
+    // Publish stuff
+    std_msgs::Int8 intmsg;
+    intmsg.data = number_of_crosses;
+
+    if (debug_pub.getNumSubscribers() > 0) {
+        // Draw some debug info
+        std::string angle_str = std::string("Angle: ") + std::to_string(angle);
+        cv::putText(debugDrawing, angle_str, cv::Point(5, 100), cv::FONT_HERSHEY_SIMPLEX, 2, cv::Scalar(0, 143, 143),
```
Daniel-Martin576: Probably should check that `fitRect` is initialized when you are printing angle and stuff

rr_iarrc/src/finish_line_watcher/finish_line_watcher.cpp
```diff
@@ -78,23 +84,122 @@ void ImageCB(const sensor_msgs::ImageConstPtr& msg) {
...
+    // Matrix used for detecting contours
+    Mat contourImg;
+    // Matrix used to draw debug info
+    Mat debugDrawing = Mat::zeros(frame.size(), CV_8UC3);
+    cv::cvtColor(frame, debugDrawing, cv::COLOR_GRAY2BGR);
+
+    // Find contours
+    vector<vector<Point> > contours;
+    vector<Vec4i> hierarchy;
+    cv::Canny(frame, contourImg, canny_thresh1, canny_thresh2, 3);
```
Daniel-Martin576: Still confused why you did Canny here instead of going straight to findContours.

rr_iarrc/src/finish_line_watcher/finish_line_watcher.cpp
```diff
@@ -13,6 +13,7 @@
 #include <sensor_msgs/Image.h>
 #include <std_msgs/Int8.h>
 #include <climits>
+#include <cmath>
 #include <opencv2/opencv.hpp>
 
 using namespace std;
```
Daniel-Martin576: Also, I know this is a bit tedious but can you remove the usage of the namespace for ros and opencv. We attempt to at most use `using namespace std;` in our codebase.

rr_iarrc/src/finish_line_watcher/finish_line_watcher.cpp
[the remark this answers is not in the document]
dsiegler2000: Good point, it was included in the tutorials for contour detection but isn't actually needed in this case.

Daniel-Martin576:
Also, press `Resolve converstation` if you completely addressed the issue and there won't be more discussion on that part so it the PR looks cleaner. Sorry for all the change requests, the code looks good honestly, just want to make sure it is in tip-top shape.

rr_iarrc/src/finish_line_watcher/finish_line_watcher.cpp
dsiegler2000: [here](https://docs.opencv.org/2.4/doc/tutorials/imgproc/shapedescriptors/find_contours/find_contours.html) , but it isnt needed and removed now and works the same

rr_iarrc/src/finish_line_watcher/finish_line_watcher.cpp
Daniel-Martin576: You should try to combine the two block with the `incrementCrossNum = True && publish_when_detected` and `incrementCrossNum = True && !publish_when_detected` trick.

rr_iarrc/src/finish_line_watcher/finish_line_watcher.cpp
Daniel-Martin576: You already have a binary image so you don't need it

rr_iarrc/src/finish_line_watcher/finish_line_watcher.cpp
```diff
@@ -41,32 +63,10 @@ void blockEnvironment(const cv::Mat& img) {
...
-    }
-    return width;
-}
-
 void ImageCB(const sensor_msgs::ImageConstPtr& msg) {
     cv_bridge::CvImagePtr cv_ptr;
-    Mat frame;
-    Mat output;
+    cv::Mat frame;
+    cv::Mat output;
```
Daniel-Martin576: I don't think `output` is used

rr_iarrc/src/finish_line_watcher/finish_line_watcher.cpp
```diff
@@ -78,30 +78,122 @@ void ImageCB(const sensor_msgs::ImageConstPtr& msg) {
...
     blockEnvironment(frame);
 
-    auto count = countNonZero(frame);
+    // Matrix used to draw debug info
+    cv::Mat debugDrawing = cv::Mat::zeros(frame.size(), CV_8UC3);
+    cv::cvtColor(frame, debugDrawing, cv::COLOR_GRAY2BGR);
+
+    // Find contours
+    vector<vector<cv::Point>> contours;
+    vector<cv::Vec4i> hierarchy;
```
Daniel-Martin576: `hierarchy` isn't used either

rr_iarrc/src/finish_line_watcher/finish_line_watcher.cpp
```diff
@@ -78,30 +78,122 @@ void ImageCB(const sensor_msgs::ImageConstPtr& msg) {
...
+        cv::Point2f rectPoints[4];
+        fitRect.points(rectPoints);
+
+        for (int i = 0; i < 4; i++)
+            cv::line(debugDrawing, rectPoints[i], rectPoints[(i + 1) % 4], cv::Scalar(255, 0, 0), 2, 12);
+    }
+
+    // Quick count of pixels as a final sanity check
+    auto count = cv::countNonZero(frame);
+    detected = detected && count > count_thresh;
```
Daniel-Martin576: Not really sure this is really needed since you are `countNonZero` on the original frame without filtering anyways.

rr_iarrc/src/finish_line_watcher/finish_line_watcher.cpp
```diff
@@ -78,30 +78,122 @@ void ImageCB(const sensor_msgs::ImageConstPtr& msg) {
...
+    // number of crosses based on whether we want to publish when detected or when we lose detection
+    if (state == LOW && detected && cooldown == 0) {
+        incrementCrossNum = publish_when_detected;
         state = HIGH;
-    } else if (state == HIGH && count < 2000) {
+
+    } else if (state == HIGH && !detected) {
+        incrementCrossNum = !publish_when_detected;
         state = LOW;
+        cooldown = cooldown_value;
```
Daniel-Martin576: I don't this you need this. You already set it later I think. Could be wrong

rr_iarrc/src/finish_line_watcher/finish_line_watcher.cpp
dsiegler2000: Do you mean with the contour stuff?

Daniel-Martin576: yeah

dsiegler2000: good point I just removed it because I had the threshold uselessly low

rr_iarrc/src/finish_line_watcher/finish_line_watcher.cpp
```diff
@@ -78,30 +74,118 @@ void ImageCB(const sensor_msgs::ImageConstPtr& msg) {
...
+
+    if (filteredContourPoints.size() > 0) {
+        // Fit a rect to the points and detect based on angle and width
+        fitRect = cv::minAreaRect(filteredContourPoints);
+        size = fitRect.size;
+        angle = std::abs(std::abs(fitRect.angle) - 90);
+        drawRectDebug = true;
+        // Check the detection criteria (must look at both width and height because OpenCV is inconsistent about how it
+        // assigns them)
+        detected = angle < angle_cutoff && (size.width > width_cutoff || size.height > width_cutoff);
```
Daniel-Martin576: If height > width and height > width_cutoff, angle should be around 0 and when width > height and width > width_cutoff, angle should be around 90. Get the angle of the major axis. Just realized this while testing on the circuit bag file.

rr_iarrc/launch/perception/finish_line_watcher.launch
```diff
@@ -0,0 +1,40 @@
...
+      <!-- Params to control blocking of the image. (Image Height = 1280)-->
+      <!-- Sky height draws the a block from y=0 to y=sky height (increasing sky height lowers the sky rectangle) -->
+      <param name="blockSky_height" type="int" value="560" />
+      <!-- Wheels height draws the a block from the bottom to y=wheel height (increasing wheel height brings the rectangle up bigger from the bottom) -->
+      <param name="blockWheels_height" type="int" value="750" />
+      <!-- Same thing as wheel height, just draws the rectangle over the middle third of the screen -->
+      <param name="blockBumper_height" type="int" value="720" />
+
+      <!--If set to true, the detector will start publishing once the finish line is first detected, if false, publishes when we cross the finish line-->
+      <param name="publish_when_detected" type="boolean" value="false" />
```
Daniel-Martin576: Should be bool

rr_iarrc/launch/perception/finish_line_watcher.launch
```diff
@@ -0,0 +1,40 @@
...
+
+      <!-- Number of camera frames between detections. Added to prevent flickering and double detections. -->
+      <param name="cooldown" type="int" value="7" />
+
+      <!-- The detector fits a bounding box to the contours in the image. min_contour_area defines the minimum area cutoff for a contour in order for it to not be excluded from this bounding box using OpenCV's contourArea units -->
+      <param name="min_contour_area" type="double" value="5.0" />
+	  <!-- In order for the fitted box to be considered a finish line, it must have an angle < angle_cutoff (in degrees) and a width > height_cutoff (in pixels) -->
+      <param name="angle_cutoff" type="double" value="20.0" />
+      <param name="height_cutoff" type="double" value="275.0" />
```
Daniel-Martin576: You say width_cutoff in main

rr_iarrc/src/finish_line_watcher/finish_line_watcher.cpp
dsiegler2000: Good point, width and height will randomly flip which is why it appeared to work.

rr_iarrc/src/finish_line_watcher/finish_line_watcher.cpp
```diff
@@ -13,21 +13,40 @@
...
 
+// Debug publisher for my image processing stuff
+ros::Publisher debug_pub;
+cv_bridge::CvImage debug_img;
+
+// Define the cutoff angle and width to detect the line and the cutoff area for contours
+double angle_cutoff;
+double width_cutoff;
+double min_contour_area;
+double area_cutoff;
```
Daniel-Martin576: This is never used.

rr_iarrc/launch/perception/finish_line_watcher.launch
Daniel-Martin576: Change `type=boolean` to `type=bool`

rr_iarrc/src/finish_line_watcher/finish_line_watcher.cpp
```diff
@@ -78,30 +74,118 @@ void ImageCB(const sensor_msgs::ImageConstPtr& msg) {
...
+        // Fit a rect to the points and detect based on angle and width
+        fitRect = cv::minAreaRect(filteredContourPoints);
+        size = fitRect.size;
+        drawRectDebug = true;
+        // Check the detection criteria (must look at both width and height because OpenCV is inconsistent about how it
+        // assigns them)
+        if (size.height > size.width)
+            angle = std::abs(std::abs(fitRect.angle) - 90);
+
+        detected = angle < angle_cutoff && (size.width > width_cutoff || size.height > width_cutoff);
```
Daniel-Martin576: Sometimes you get -0 when you skip the if-block so just do `std::abs(angle) < angle_cutoff`

rr_iarrc/src/finish_line_watcher/finish_line_watcher.cpp
```diff
@@ -110,21 +194,20 @@ int main(int argc, char** argv) {
...
+    nhp.param("angle_cutoff", angle_cutoff, 20.0);
+    nhp.param("width_cutoff", width_cutoff, 275.0);
+
     ROS_INFO("Finish line watching %s", img_topic.c_str());
 
-    Subscriber img_saver_sub = nh.subscribe(img_topic, 1, ImageCB);
+    ros::Subscriber img_saver_sub = nh.subscribe(img_topic, 1, ImageCB);
 
     crosses_pub = nh.advertise<std_msgs::Int8>("finish_line_crosses", 1);
+    debug_pub = nh.advertise<sensor_msgs::Image>("finish_line_detection/debug", 1);
```
Daniel-Martin576: Change it to this `nhp.advertise<sensor_msgs::Image>("debug", 1);` so it has the same names as the node

rr_iarrc/src/finish_line_watcher/finish_line_watcher.cpp
```diff
@@ -110,21 +193,20 @@ int main(int argc, char** argv) {
...
+    nhp.param("angle_cutoff", angle_cutoff, 20.0);
+    nhp.param("width_cutoff", width_cutoff, 275.0);
+
     ROS_INFO("Finish line watching %s", img_topic.c_str());
 
-    Subscriber img_saver_sub = nh.subscribe(img_topic, 1, ImageCB);
+    ros::Subscriber img_saver_sub = nh.subscribe(img_topic, 1, ImageCB);
 
     crosses_pub = nh.advertise<std_msgs::Int8>("finish_line_crosses", 1);
+    debug_pub = nh.advertise<sensor_msgs::Image>("debug", 1);
```
Daniel-Martin576: Use `nhp` instead of `nh` to use private namespace (aka the node name)

dsiegler2000: Should I change the subscriber to use `nph` too and remove the `nh` variable ?

Daniel-Martin576:
Also, can you also remove all references to `race_end_detector` in the repo. Just do `grep -r "race_end_detector"`. This fine_line_watcher launch file is replacing it. It _should_ only be in the event launch files

rr_iarrc/src/finish_line_watcher/finish_line_watcher.cpp
ebretl: The reason to use the private NodeHandle in this case is so that the debug image topic is prefixed by the node name (so `/<node namespace>/<node name>/debug`). This is preferred for debugging information, but not necessarily for system-level messages like whether the race is over

ebretl: Using the private NodeHandle to get the node's name means that if we change the node name (in roslaunch or through refactoring), the name of debug image's topic gets updated automatically

ebretl: So the answer to "Should I change the subscriber to use nph too and remove the nh variable?" is no in my opinion

dsiegler2000: Okay makes sense, thanks

Daniel-Martin576:
Delete the race_end_detector launch file as well

rr_iarrc/src/finish_line_watcher/finish_line_watcher.cpp
```diff
@@ -78,30 +73,118 @@ void ImageCB(const sensor_msgs::ImageConstPtr& msg) {
...
+
+    if (filteredContourPoints.size() > 0) {
+        // Fit a rect to the points and detect based on angle and width
+        fitRect = cv::minAreaRect(filteredContourPoints);
+        size = fitRect.size;
+        drawRectDebug = true;
+        // Check the detection criteria (must look at both width and height because OpenCV is inconsistent about how it
+        // assigns them)
+        if (size.height > size.width)
+            angle = std::abs(fitRect.angle) - 90;
```
Daniel-Martin576: Where do you update `angle` when this is skipped? I think you are leaving it un-initialized

Daniel-Martin576: That should be the last fix

rr_iarrc/src/finish_line_watcher/finish_line_watcher.cpp
```diff
@@ -78,30 +73,118 @@ void ImageCB(const sensor_msgs::ImageConstPtr& msg) {
...
-    init(argc, argv, "finish_line_watcher");
+    ros::init(argc, argv, "finish_line_watcher");
 
-    NodeHandle nh;
-    NodeHandle nhp("~");
+    ros::NodeHandle nh;
+    ros::NodeHandle nhp("~");
 
     string img_topic;
     nhp.getParam("img_topic", img_topic);
```
oswinso: I think it would be nice to use the `assertions` helper functions to make sure that these parameters are set

Daniel-Martin576: Honestly, the only places `assertions` are really used in our codebase is in Evan's code and we should probably use them more, but I wasn't going to strictly push it in this PR. I will leave it up to Dylan's discretion.

PR #292, opened 2019-12-03
Color filter library

### LINKED ISSUE #290: Reconfigurable color filter library

ebretl:
A little class that knows how to run cv::inRange on a rectangular region of an image (in multiple color spaces) and how to reconfigure itself. I think this will be useful for the cone detectors, the finish line detector, and especially the (Laplacian) line detector which has multiple colors to filter out. This is designed to replace the HSV tuner node for the IARRC color detector.

You can test the code by running `rr_color_filter_test_node` in `rr_common` and running Macaroni in the simulator. 

Closes #290


Daniel-Martin576:
I really like this, it works really well. I think it would be helpful it there was options to:
- resize the image beforehand (WxH), usually helps with performance, will have to convert the ROI to new values
- Maybe the ability to publish just the ROI or the full image 
- resize the image window (not really necessary, opencv doesn't play nice when you are updating the image in a window for some reason)

Daniel-Martin576:
Also, something I have to remove the car's hood from the image as well as the sky and the wheels. The ROI can deal with the sky and the wheels, but maybe we can add an adjustable black box at the bottom of the image to eliminate the hood.

PR #293, opened 2019-12-08
Costmap Planning (#285)

### LINKED ISSUE #285: Update Planner to take CostMap Input
EVGP car should make a decision based on the local map and not just the current point cloud so update the planner to take in a local inflation layer costmap.

ebretl:
Fixes #285 

TBH most of the code changes here are refactoring the path planner.


Daniel-Martin576:
Make sure to update all the IARRC config files so that they still work

Daniel-Martin576:
It seems like points on the car are being added to the map. Where do we do that check?

Daniel-Martin576:
Lol, the IARRC car really zooms in simulation now

Daniel-Martin576:
I am getting a lot of `[/planner] linear tracking model found jump backwards in time 190.360000 190.360000`, I think you might have forgotten an equals sign when testing if the time goes backward

rr_common/include/rr_common/planning/map_cost_interface.h
```diff
@@ -0,0 +1,52 @@
...
+     * @param poses (x, y, theta) relative to the current pose of the robot
+     * @return for each entry, distance cost if not in collision, negative value if in collision
+     */
+    virtual std::vector<double> DistanceCost(const std::vector<Pose>& poses) = 0;
+    virtual std::vector<double> DistanceCost(const std::vector<PathPoint>& path_points) = 0;
+
+    virtual bool IsMapUpdated() {
+        return updated_;
+    }
+    virtual void SetMapStale() {
```
Daniel-Martin576: SetMapState?

ebretl: I'm using "stale" as the opposite of "recently updated"

ebretl:
> It seems like points on the EVGP car are being added to the map. Where do we do that check?

I think that would be the robot footprint/radius in the mapper config (common params I think). Also check that footprint clearing is enabled

rr_common/src/planner/planner_node.cpp
```diff
@@ -226,27 +222,35 @@ int main(int argc, char** argv) {
...
     ros::Rate rate(30);
-    is_new_msg = false;
     while (ros::ok()) {
+        rate.sleep();
         ros::spinOnce();
 
-        if (is_new_msg) {
-            is_new_msg = false;
+        g_steer_model->Update(steer_message->angle, ros::Time::now().toSec());
+        g_speed_model->Update(speed_message->speed, ros::Time::now().toSec());
```
Daniel-Martin576: Shouldn't we update these with our actual current speed and steering, not the theoretical?

rr_evgp/conf/mapping_local.yaml
```diff
@@ -1,57 +1,35 @@
...
-        max_dist_between_lines: 50
+        extrapolate_distance: 1
+        max_angle_between_lines: 30
+        max_dist_between_lines: 20
         wall_output_dilation: 1
+
+    distance_field:
+        inflation_radius: 35
+        cost_scaling_factor: 0.4
+        lethal_cost_threshold: 255
```
Daniel-Martin576: I don't think this is needed:
`The threshold value at which to consider a cost lethal when reading in a map from the map server.`

rr_common/src/planner/planner_node.cpp
ebretl: Perhaps that should be done with the steering model, for platforms where we have steering angle feedback. (To my knowledge, we don't have steering angle feedback on any of our bots yet.) 

Also, the speed model is also being used as a filter on the speed that is published by the node (planner_node.cpp line 139). So the value of the speed model is more of the ideal speed than the predicted speed. I'm actually not sure that line 238 is necessary.

ebretl:
> I am getting a lot of [/planner] linear tracking model found jump backwards in time 190.360000 190.360000, I think you might have forgotten an equals sign when testing if the time goes backward

Fixed

rr_evgp/conf/mapping_local.yaml
ebretl: Good point, just checked the source and that param is not used. idk why I included that

rr_common/src/planner/planner_node.cpp
Daniel-Martin576: Don't we know our speed due to localization for evgp? The main reason I say this is because when we are running the path planner and forcing the speed to 0 (like on start light) then the speed will ramp up pretty high right when the light goes green.

Daniel-Martin576: I was testing stuff and forcing the speed to 0 and the paths started fairly straight, then the lines were acting weird (very very wobbly) as the speed reached the max and the real speed was still 0.

ebretl: It makes sense to me that there would be serious problems when the planner thinks the vehicle is going way faster or slower than it really is. What if we reset the speed model to the real value whenever we get an update from /chassis_state, and we use the model's predictions in between?

Daniel-Martin576: I was thinking of using the odometry msg from robot localization that mixes all our sensors. 
I believe that chassis_state will only give encoder data. We can just make it an optional parameter thou in the planner

ebretl: Right, that would be better

PR #289, opened 2019-12-01
Feat/demo follow

erclin:
Fix #276 
Implemented PID using the pid package for the demo follower. Also changed it from using the average error to minimum error.

PR #294, opened 2019-12-08
Feat/demo follow

### LINKED ISSUE #276: Clean demonstration follower
**Goal:** Update the current `follower.cpp` code to use a PID controller to reach a target distance away from the closest point detected from lidar. Meaning that if a person put their hand in front of the car's lidar and began moving their hand back and forth, then the car would attempt to remain a set distance away from it. 

**Notes:** The current code oscillates back and forth quite a bit since it is simply attempting to stop within some margin of error of the target distance using a constant forward and backward velocity. Either implement a PID controller similar to the week3 ros-training exercise or use the [PID](http://wiki.ros.org/pid) ros package to control the speed of the car. The ros package implementation is if you want to go above and beyond. 
Test the behavior of the controller by launching a simulator by running `roslaunch rr_gazebo macaroni_avc.launch` and moving newly-created cubes in front of the car. 
Related to issue #208 

**Files Edited (at least):** `/rr_common/src/follower/follower.cpp`, `/rr_common/launch/demo_obstacle_follower.launch`

**Returning Member Mentor:** Daniel Martin @Daniel-Martin576  --Ask him questions!
When you believe you have finished the success criteria below, submit a [pull request](https://help.github.com/en/articles/about-pull-requests) to merge your code into master.

**Success Criteria:**
- Able to launch demonstration follower mode
- The car moves a set distance away from the closest lidar point and stops there without much oscillation
- Behavior has been tested in simulation (Gazebo world)

**Future Work: (if you want to give it a try, but don't waste too much time on it)**
- Use a [PID ros package](http://wiki.ros.org/pid) in order to maintain set distance from the closest lidar point.

erclin:
Improves #289 which fixed #276 
Added a visual representation of the bounding box. Also changes the launch variables for the pid topics into arguments.

### DIFF
--- a/rr_common/launch/demo_obstacle_follower.launch
+++ b/rr_common/launch/demo_obstacle_follower.launch
@@ -1,14 +1,20 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <launch>
+    <arg name="topic_from_controller" value="/control_effort"/>
+    <arg name="topic_from_plant" value="/state"/>
+    <arg name="setpoint_topic" value="/setpoint"/>
     <node name="follower" pkg="rr_common" type="follower" output="screen">
-        <param name="MIN_FRONT_VISION" value="0.2"/>
-        <param name="MAX_FRONT_VISION" value="5"/>
+        <param name="MIN_FRONT_VISION" value="0.37"/>
+        <param name="MAX_FRONT_VISION" value="3"/>
         <param name="GOAL_DIST" value="1.3"/>
         <param name="GOAL_MARGIN_OF_ERR" value="0.2"/>
         <param name="MIN_SIDE_VISION" value="-0.5"/>
         <param name="MAX_SIDE_VISION" value="0.5"/>
         <param name="FOLLOWER_SPEED" value="0.2"/>
         <param name="INPUT_CLOUD_TOPIC" value="/scan/pointcloud"/>
+        <param name="topic_from_controller" value="$(arg topic_from_controller)" />
+        <param name="topic_from_plant" value="$(arg topic_from_plant)" />
+        <param name="setpoint_topic" value="$(arg setpoint_topic)" />
     </node>
     <node name="demo_pid_controller" pkg="pid" type="controller" >
         <param name="Kp" value="-2.0" />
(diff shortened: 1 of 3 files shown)


rr_common/launch/demo_obstacle_follower.launch
```diff
@@ -1,14 +1,20 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <launch>
+    <arg name="topic_from_controller" value="/control_effort"/>
+    <arg name="topic_from_plant" value="/state"/>
+    <arg name="setpoint_topic" value="/setpoint"/>
```
Daniel-Martin576: Add `follower/` to the front of the topic names so that these topics won't interfere with other launch files where we also use the pid package.

rr_common/src/follower/follower.cpp
```diff
@@ -92,6 +106,8 @@ int main(int argc, char** argv) {
     nhp.param("setpoint_topic", setpoint_topic, string("/setpoint"));
     nhp.param("topic_from_controller", topic_from_controller, string("/control_effort"));
 
+    visonBox_pub = nh.advertise<geometry_msgs::PolygonStamped>("vision_box", 1);
```
Daniel-Martin576: Also, add `follower/` here either by just adding `follower/vision_box` topic name with nh or just `vision_box` with nhp which is the node's private node handle.

rr_common/src/follower/follower.cpp
```diff
@@ -92,6 +106,8 @@ int main(int argc, char** argv) {
     nhp.param("setpoint_topic", setpoint_topic, string("/setpoint"));
     nhp.param("topic_from_controller", topic_from_controller, string("/control_effort"));
 
+    visonBox_pub = nhp.advertise<geometry_msgs::PolygonStamped>("/follower/vision_box", 1);
```
Daniel-Martin576: If you are using `nhp`, ros includes the node name (in this case /follower) in front of the topic name if you don't include a slash (/) in front of the name. So just do `nhp.advertise<geometry_msgs::PolygonStamped>("vision_box", 1);` so that the topic name changes if we change the node name.

PR #307, opened 2019-12-30
Adds proposed 2020 EVGP track

btdubs:
This is the proposed track for EVGP 2020 in sim based on https://www.ekseries.com/post/sneak-peak-of-the-2020-world-final-track-design-10370129

Closes #296 

The old track is still available as EVGP holes because it features gaps in the walls.

My simulation lidar is not working (even on the old track) so someone should test this on their machine.

PR #309, opened 2020-01-02
Planner Calculates its Own Distance Cost Map

### LINKED ISSUE #306: Create Distance Map Innerly with Planner
Allow the option to run the planner where the distance map is calculated internally instead of using the inflation layer. 
Test performance benefit.

Daniel-Martin576:
Similar to the inflation cost map, but now we have more control over the cost function. 
Distance Cost Map includes:
- Wall Inflation 
- Inflates collision area to include the footprint's inscribed circle (similar to inflation layer)
- Publishes generated cost map
- Shifted forwards the origin of the inscribed circle check so that half of the circle isn't outside the hitbox. 
- Published inscribed circle
- Made car footprint rectangular and extended the back so that we would catch fewer points on the car

Fixes #306 
Make sure to merge evgp proposed track PR #307 first. I used it for testing.


rr_common/src/planner/distance_map.cpp
```diff
@@ -0,0 +1,103 @@
...
+}
+
+void DistanceMap::SetMapMessage(const boost::shared_ptr<nav_msgs::OccupancyGrid const>& map_msg) {
+    if (!accepting_updates_) {
+        return;
+    }
+
+    try {
+        listener->waitForTransform(map_msg->header.frame_id, "base_footprint", ros::Time(0), ros::Duration(.05));
+        listener->lookupTransform(map_msg->header.frame_id, "base_footprint", ros::Time(0), transform);
```
ebretl: `base_footprint` should probably be replaced by a param

rr_common/src/planner/distance_map.cpp
```diff
@@ -0,0 +1,103 @@
...
+    } catch (tf::TransformException& ex) {
+        ROS_ERROR_STREAM(ex.what());
+    }
+
+    mapMetaData = map_msg->info;
+
+    // Turn occupancy grid to distance map in meters
+    cv::Mat distance_map(mapMetaData.width, mapMetaData.height, CV_8UC1);
+    memcpy(distance_map.data, map_msg->data.data(), map_msg->data.size() * sizeof(uint8_t));
```
ebretl: `memcpy` smells like a potential bug, but if the memory layouts of the message and the Mat are the same, then I guess it's fine

Daniel-Martin576: Both are continuous so it works fine, but I can change it to the nested for-loop way if you think `memcpy` will produce problems.

PR #310, opened 2020-01-03
Added Speed and Steering Tracker

### LINKED ISSUE #305: Update the Planner to take in Current Speed
Update path planner to subscribe to optional localization odometry message in order to update the initial speed for each planned path. 
```c++
g_steer_model->Update(steer_message->angle, ros::Time::now().toSec());
g_speed_model->Update(speed_message->speed, ros::Time::now().toSec());
```
Probably should be done for the steering model as well, but currently we don't have steering angle feedback
Allow for two modes: one where the initial speed is locked on the last odometry data until the next message is received or let the planner predict what the current speed is in between messages based on what it just published.

Daniel-Martin576:
Fixes #305 
Now our planner knows our current speed and steering angle.
Given a message topic and type it can:
- Extract speed from chassis state, odometry, and  the last planner speed message 
- Extract steering angle from chassis state and the last planner steering message
 
You can let it guess what the current speed and steering angle is just by allowing it to reuse the previously published planner message.


rr_common/src/planner/effector_tracker.cpp
```diff
@@ -0,0 +1,80 @@
+#include "rr_common/planning/effector_tracker.h"
+
+namespace rr {
+
+EffectorTracker::EffectorTracker(ros::NodeHandle nh, rr_msgs::speedPtr& speed_msg, rr_msgs::steeringPtr& steering_msg)
```
dsiegler2000: Would be nice to have some type of documentation comment

PR #314, opened 2020-02-04
Feat/Pitch Roll Estimator

### LINKED ISSUE #302: Pitch and Roll Estimate
Pitch, roll estimate for point projector for Sedanii -- test accuracy of the current projector

Our point projector takes in a camera image (usually a binary image of the lines detected) and maps every pixel in the image to a point in a point cloud (so that pixels higher in the image are farther away and vice versa). This works okay when the car is steady since we assume the camera is at a fixed angle; however, when our car jerks forward and backward during sudden accelerations the projected points will quickly appear really close then super far away. We need a pitch and roll estimator so that we know the current position of the camera to correct this.

dsiegler2000:
Closes #302 

Adds functionality to improve our pointcloud projector. Uses the IMU to read in the axes and updates the camera pose to reflect the rotation of the chassis (pitch and roll primarily due to the suspension moving). These updates are propagated through the pointcloud projector.

### DIFF
--- a/rr_common/src/image_transformation/pointcloud_projector.cpp
+++ b/rr_common/src/image_transformation/pointcloud_projector.cpp
@@ -11,9 +11,9 @@
 #include <tf/transform_listener.h>
 #include <tf2_geometry_msgs/tf2_geometry_msgs.h>
 
+#include <geometry_msgs/Pose.h>
 #include <sensor_msgs/Image.h>
 #include <sensor_msgs/PointCloud2.h>
-#include <geometry_msgs/Pose.h>
 
 #include <rr_common/CameraGeometry.h>
 #include <rr_msgs/axes.h>
@@ -23,12 +23,8 @@
 using PointT = pcl::PointXYZ;
 
 ros::Subscriber axes_sub_;
-tf::TransformListener tf_listener;
-tf::StampedTransform camera_optical_center_tf;
-
-// Pitch and roll variables (initialized to 0 for the start)
-geometry_msgs::Pose camera_pose;
-
+bool axes_init;  // Stores whether the initial axes have been added
+tf2::Quaternion initial_axes;
 
 class PointCloudProjector : public nodelet::Nodelet {
   private:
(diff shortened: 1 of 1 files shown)


rr_common/src/image_transformation/pointcloud_projector.cpp
```diff
@@ -68,10 +78,34 @@ class PointCloudProjector : public nodelet::Nodelet {
...
+        tf2::Quaternion quat_tf;
+        tf2::convert(cam_geom_.GetCameraPose().orientation, quat_tf);
+
+        // Set roll, pitch, and yaw
+        quat_tf.setRPY(ax.roll, ax.pitch, ax.yaw);
+        // Calculate the DELTA of the pitch and roll
+        quat_tf -= initial_axes;
+
+        // Convert back
+        cam_geom_.GetCameraPose().orientation = tf2::toMsg(quat_tf);
```
ebretl: GetCameraPose returns a Pose by value, so changing this orientation does not change the internal state of cam_geom_

rr_common/src/image_transformation/pointcloud_projector.cpp
```diff
@@ -68,10 +78,34 @@ class PointCloudProjector : public nodelet::Nodelet {
         pointcloud_pub_.publish(out);
     }
 
+    void axes_callback(const rr_msgs::axes& ax) {
+        if (!axes_init) {
+            // Set the initial axes
+            tf2::convert(cam_geom_.GetCameraPose().orientation, initial_axes);
+        }
```
ebretl: Does this assume that the car is upright and flat when the node is launched? That's fine as long as it's documented

ebretl:
I think this class in a ROS package does the 3D geometry stuff right, but like my CameraGeometry class it also only updates through TF which might be too slow for "real-time" reactions to our pitch and roll. http://docs.ros.org/melodic/api/image_geometry/html/c++/classimage__geometry_1_1PinholeCameraModel.html

Here is my recommendation: fix CameraGeometry so that a user can change its orientation and/or pose, but continue to ignore the roll dimension in ProjectToWorld. I would expect more of our present error to come from ignoring pitch than from ignoring roll. Then proper 3D point projection (simplified raycasting) can be added in later if it's still needed

Daniel-Martin576:
Going to close this for now (a bit stale) and link it in the issue.

PR #318, opened 2020-04-07
Global Center Line

### LINKED ISSUE #273: Global Center Line Finder
Find the centerline of the track after the cold lap. It should be a list of points that produce a closed circuit around the track through the middle of the lane. 

Steps:
- Find the distance of each cell to the closest wall
     - Convert to image and use OpenCV distance transform
     - Try to find something else online
     - Outside the track should be 0 and inside should be some positive x distance
- Use Uniform-cost Search to find loop around the course using the distance as a utility  
     - Set goal point behind you and add a wall between the start point and goal point to prevent the search from going backward
- Smooth the resultant circuit if needed 

Other Option (probably try this first):
- Just use navfn and global_planner packages and see how good a result you get

btdubs:
Closes #273 by adding a costmap layer for calculating the center line of the map.

The procedure is to use the free_space as the search space, calculate distance from free cells to obstacles, convert distance to cost by setting highest distance to 0 increasing cost the lower the distance, then uniform cost search.

This did adjust the costmap to be trinary state which adjusted the distance_map cost calculations. I have not yet tested it with out with the other map cost interfaces to make sure they work with trinary state maps.

Down scaling the map speeds search time, at the cost of resolution of path. Future improvements could be made to smooth the path, but I wanted to get this out and work on adding path costs to the planner first.


rr_evgp/src/UniformCostSearch/UniformCostSearch.cpp
```diff
@@ -0,0 +1,126 @@
...
+    s.pt = startPoint_;
+    s.cost = 0.0;
+    return s;
+}
+
+bool UniformCostSearch::isGoalState(UniformCostSearch::State s) {
+    if (s.pt == goalPoint_) {
+        return true;
+    }
+    return false;
```
Daniel-Martin576, line 27: Could be `return s.pt == goalPoint_` but it honestly doesn't matter.

rr_evgp/src/UniformCostSearch/UniformCostSearch.cpp
```diff
@@ -0,0 +1,126 @@
...
+}
+
+float UniformCostSearch::pointToCost(cv::Point pt) {
+    return distanceGrid_.at<float>(pt);  // DistanceTransform outputs CV_32F
+}
+
+bool UniformCostSearch::isValidPoint(cv::Point pt) {
+    cv::Rect rect(cv::Point(), distanceGrid_.size());
```
Daniel-Martin576, line 35: Don't think it would make much of a difference but this could be a global variable or static so you didn't have to create it for every call

rr_evgp/src/UniformCostSearch/UniformCostSearch.cpp
```diff
@@ -0,0 +1,126 @@
...
+    // returns the 8 point grid around a center point
+    std::vector<UniformCostSearch::State> successors;
+    for (int i = -1; i < 2; i++) {
+        for (int j = -1; j < 2; j++) {
+            if (i != 0 || j != 0) {  // not center position
+                cv::Point pt(s.pt.x + i, s.pt.y + j);
+                if (this->isValidPoint(pt)) {
+                    UniformCostSearch::State newState;
+                    newState.pt = pt;
+                    newState.cost = this->pointToCost(pt);
```
Daniel-Martin576, line 49: Not sure why you do `this->function()` instead of just `function()`, I think you do this some other places too

rr_evgp/src/UniformCostSearch/UniformCostSearch.cpp
```diff
@@ -0,0 +1,126 @@
...
+    std::priority_queue<UniformCostSearch::State, std::vector<UniformCostSearch::State>,
+                        std::greater<UniformCostSearch::State>>
+          pq;  // minimum priority queue
+    cv::Mat visited(obstacleGrid_.size(), CV_8UC1, cv::Scalar(0));
+
+    pq.push(this->getStartState());
+    // uniform cost search
+    while (!pq.empty()) {
+        UniformCostSearch::State e = pq.top();
+        if (visited.at<uchar>(e.pt) == 0) {
```
Daniel-Martin576, line 68: `== FREE_SPACE`?

rr_evgp/src/UniformCostSearch/UniformCostSearch.cpp
```diff
@@ -0,0 +1,126 @@
...
+            if (this->isGoalState(e)) {
+                return e.path;
+            }
+            std::vector<UniformCostSearch::State> successors = this->getSuccessors(e);
```
Daniel-Martin576, line 72: Could just return `vector<cv::Point>` and just get state cost when adding to parent state cost. Both work fine

rr_evgp/src/UniformCostSearch/UniformCostSearch.cpp
```diff
@@ -0,0 +1,126 @@
...
+            for (UniformCostSearch::State s : successors) {
+                UniformCostSearch::State newState;
+                newState.pt = s.pt;
+                newState.path = e.path;
+                newState.path.push_back(s.pt);
+                newState.cost = e.cost + s.cost;
+                pq.push(newState);
+            }
+            visited.at<uchar>(e.pt) = 255;
```
Daniel-Martin576, line 81: `== OBSTACLE_SPACE`?

rr_evgp/src/costmap_plugins/global_center_path_layer.cpp
```diff
@@ -0,0 +1,235 @@
...
+                double min;
+                double max;
+                cv::minMaxLoc(distanceGrid, &min, &max);
+                cv::absdiff(distanceGrid, cv::Scalar(max), distanceGrid);
+                cv::bitwise_not(obstacleGrid, obstacleGrid);
+
+                cv::Point startPt(mx / scaling_factor_, my / scaling_factor_);
+                goalPt.x = goalPt.x / scaling_factor_;
+                goalPt.y = goalPt.y / scaling_factor_;
+                UniformCostSearch ucs(obstacleGrid, distanceGrid, startPt, goalPt);
```
Daniel-Martin576, line 164: Could just add this after you find the next(Start/Goal)Pt.

rr_evgp/src/costmap_plugins/global_center_path_layer.cpp
```diff
@@ -0,0 +1,235 @@
...
+                cv::dilate(obstacleGrid, obstacleGrid, kernel);
+
+                cv::bitwise_not(obstacleGrid, obstacleGrid);  // Flip due to how DistanceTransform defines obstacles
+                cv::Mat distanceGrid;
+                cv::distanceTransform(obstacleGrid, distanceGrid, cv::DIST_L2, cv::DIST_MASK_3);
+                // convert based on max so that we minimize the center path
+                double min;
+                double max;
+                cv::minMaxLoc(distanceGrid, &min, &max);
+                cv::absdiff(distanceGrid, cv::Scalar(max), distanceGrid);
```
Daniel-Martin576, line 158: Could do `cv::add` with `-max` or just do `cv::absdiff`. Not sure why there isn't a 'cv::diff` function

rr_evgp/src/costmap_plugins/global_center_path_layer.cpp
```diff
@@ -0,0 +1,235 @@
...
+                cv::bitwise_not(obstacleGrid, obstacleGrid);
```
Daniel-Martin576, line 159: Could save a `obstacleGrid_not` and normal `obstacleGrid` if you didn't want to `not` then `not`

Issue #320, opened 2020-08-25
Phantom track noetic


NicoBartholomai:
Update rosbag_to_trajectory_csv.py to run in Python 3 for Noetic support.
Create a new csv with a bag file that recorded the bot going around the new track without holes.
Test by running evgp_with_opponent.launch
see #313

PR #322, opened 2020-09-22
Feat/320 adding opponent cars

NicoBartholomai:
Fixes issue #320 
Adds multiple cars to the track when launching evgp_with_opponent.launch.
Added two new spawning nodes for two extra cars, as well as created three new csv files that determine their path.
Also created 5 new opponent cars in rr_description/urdf/opponents that each have their own color, as well as unique collision bit-masks so that the opponent cars cannot run into each other and lag out.
Adds two new parameters to util/rosbag_to_trajectory_csv.py which adds the functionality to give the kart a starting position other than 0, 0 when creating a path csv from a bag file, since ros does not track what the initial gazebo coordinates are.
Tested by running 'roslaunch rr_gazebo evgp_with_opponent.launch' immediately followed by "roslaunch rr_evgp evgp_planning_demo.launch" and then saw the three cars spawned and moved around the track.

Video of the test run can be found [here](https://drive.google.com/file/d/1MsNn4GokU0b5xXBxXJhk5zrNjnXFcs-0/view?usp=sharing)


Daniel-Martin576:
Can you update this branch so that the file changed is only the ones you actually touched? Just merge master and re-push pretty much

Daniel-Martin576 (approved):
Looks good. Nice Job! :+1: :100:

PR #324, opened 2020-10-11
Feat/323 urban simulator world

### LINKED ISSUE #323: Urban Simulator World
Create a world in Gazebo for the urban challenge in iarrc.
Must have road-like streets with sharp terms, as well as a sign at each intersection with arrows pointing where to go.
It should also have a stoplight in the beginning that tells the car when to start.
This is important for testing the current Urban Challenge stack virtually and should try to mimic the actual competition as much as possible.

NicoBartholomai:
fixes #323 
Adds a map for the urban challenge into gazebo with a stoplight that will change from red to green after a couple of seconds.
The map can be launched with ```roslaunch rr_gazebo macaroni_iarrc_urban.launch```
The map follows the path, with directional arrows, outlined [here](https://iarrc.org/wp-content/uploads/2020/01/IARRC-2020-Rules-v1.0.pdf)
Tested by running the launch file and seeing that the car spawns at the correct place and that the light changes after a couple of seconds. Video of it running can be found [here](https://drive.google.com/file/d/1wN5Ll625O_Eul_AhsfPdvbzbtNIlGH84/view?usp=sharing).


Daniel-Martin576 (approved):
Looks good!

Issue #332, opened 2020-10-31
Set nodes in hsv_tuner.launch to be required


Daniel-Martin576:
For `hsv_tuner.launch` in `rr_iarrc/launch/perception`, add the attribute `required="true"` to the node:
- hsv_tuner 

Once you make the PR, assign a fellow RoboRacing new member to review it.

PR #339, opened 2020-10-31
add required=true attribute to hsv_tuner.launch

Udit8348:
Address issue #332 by adding `required="true"` attribute to the node.


jshao109 (approved):
Looks good

PR #342, opened 2020-11-01
require finishline watcher

ruiyi-gao:
Reselves #331


wbeard01 (commented):
The finish_line_watcher.launch change looks good, but I'm not sure what's going on with the pointgrey_camera_driver submodule. Maybe @Daniel-Martin576 has some ideas.

PR #350, opened 2020-11-08
Fix side cameras

Daniel-Martin576:
Side cameras weren't set up properly for simulation, so I quickly fixed them. I tested them in the urban challenge simulation to ensure that both lanes were being detected.


NicoBartholomai (approved):
Works on my system. Seems good to me.

PR #359, opened 2020-11-08
Update Rigatoni URDF 

### LINKED ISSUE #299: Add 3D Lidar
Since RR doesn't a 3D lidar yet, IGVC was kind enough to let us borrow their nice Velodyne one for our comp! It is way nicer to have a 3D lidar for obstacle detection, so update our lidar in the gazebo to match the specs of their lidar (you can look through their repo to see how they did it) ([urdf.xacro](https://github.com/RoboJackets/igvc-software/blob/master/igvc_description/urdf/jessii.urdf.xacro) and [urdf](https://github.com/RoboJackets/igvc-software/blob/master/igvc_description/urdf/jessi.urdf)). If you could give the option to switch between the two types (2D and 3D) it would be pretty nice too. 

Right now our stack uses the 2D lidar /scan to make a map, so if you are able to produce a 2D map of the world with 3D scan it would be ideal! You can either make your own node or maybe look into `octomap` and `pcl_ros`.

### LINKED ISSUE #316: Adding April Tags to Gazebo model of the car
This issue will add april tags to the car configuration.

Daniel-Martin576:
WIP
Resolves #299 and resolves #316 by adding 3D lidar and April tags.
Also, added launch arg for switching between models. 
Still need to add April tags to opponent cars and add other IMU sensor. 

Tested in simulation 
Don't have the car running yet though since we need convert 3D to 2D lidar scan.


rr_evgp/launch/localization_sim.launch
```diff
@@ -22,7 +22,7 @@
         <param name="use_odometry_yaw" value="false"/>
         <param name="wait_for_datum" value="false"/>
 
-        <remap from="/imu/data" to="/imu"/>
+        <remap from="/imu/data" to="/imu_bottom"/>
         <remap from="/gps/fix" to="/gps/fix"/>
         <remap from="/odometry/filtered" to="/odometry/filtered"/>
```
NicoBartholomai, line 27: Do these remaps end up doing anything?

BestCharlemagne:
The AprilTags are too small to be usable. They are only 10x10 and ROS just blurs them when they are enlarged.

rr_evgp/launch/localization_sim.launch
Daniel-Martin576: We might remap them to something else in the future (if we change localization stuff) so I think we should keep it in order to remind ourselves that this node uses it.

NicoBartholomai (approved):
Looks good to me, the April-tag is much more clear now.

PR #360, opened 2020-11-10
Feat/325 dynamic time warp

### LINKED ISSUE #325: Dynamic Time Warp
Update the global path to use dtw to compare the sample path and the global path to give a better estimate of how close the two paths are to the planner. This would fix the issue where a path that circles back on itself would look very close to the global path even though it is not.

NicoBartholomai:
Closes #325 Updates the cost function to include a comparison via dynamic time warp (dtw) between the global path and the current plan. For each generated plan it finds the closest point on the global path to the bot, and then generates a global path segment about the same length as the path. It then uses dtw to compare the current plan to the global cost. 

Tested by launching evgp_sim.launch as well as evgp_planning_demo.launch and observing that the path on the topic /global_path_seg is of similar length to the projected path, and that it begins approximately where the robot begins.

Video of it running can be found [here](https://drive.google.com/file/d/1DtDQZXx0rI_jhrQhAeuoGyQqQCHHlRdr/view?usp=sharing).


rr_common/src/planner/planner_node.cpp
```diff
@@ -69,6 +71,8 @@ void processMap() {
         const auto& path = rollout.path;
 
         std::vector<double> map_costs = g_map_cost_interface->DistanceCost(path);
+        double global_path_costs = g_global_path_cost->CalculateCost(path, false);
+        //        double global_path_costs = 0;
```
Daniel-Martin576, line 75: Probably should remove this

Daniel-Martin576 (changes requested):
Looks good, just some stuff that should be changed a bit.

rr_common/src/planner/global_path.cpp
```diff
@@ -0,0 +1,172 @@
...
+        global_segment.insert(global_segment.end(), global_path_.begin(), global_path_.begin() + seg_end_index);
+    }
+    return global_segment;
+}
+
+double GlobalPath::dtw_distance(const std::vector<tf::Point> &path1, const std::vector<tf::Point> &path2, int w) {
+    int n = path1.size();
+    int m = path2.size();
+
+    std::vector<std::vector<double>> dtw(n + 1, std::vector<double>(m + 1, INFINITY));
```
Daniel-Martin576, line 101: Use `std::numeric_limits<double>::infinity()`

rr_common/src/planner/global_path.cpp
```diff
@@ -0,0 +1,172 @@
...
+    w = std::max(w, std::abs(n - m));  // adapt window size
+    dtw[0][0] = 0;
+
+    for (int i = 1; i < n + 1; i++) {
+        for (int j = std::max(1, i - w); j < std::min(m, i + w) + 1; j++) {
+            dtw[i][j] = 0;
+        }
+    }
```
Daniel-Martin576, line 109: I still think you don't need to put 0 for the entire part of the matrix we will be exploring.

rr_common/src/planner/global_path.cpp
```diff
@@ -0,0 +1,172 @@
...
+    std::vector<tf::Point> global_segment = GlobalPath::get_global_segment(sample_path);
+
+    // use dtw to comp sample to global
+    int window = (int)(dtw_window_factor_ * std::max(global_segment.size(), sample_path.size()));
+    double dtw_val = GlobalPath::dtw_distance(global_segment, sample_path, window);
+
+    // for publishing / debugging
+    if (viz) {
+        nav_msgs::Path global_seg_msg;  // convert type
+        for (auto path_point : global_segment) {
```
Daniel-Martin576, line 52: Const reference

rr_common/src/planner/global_path.cpp
```diff
@@ -0,0 +1,172 @@
...
+        global_segment =
+              std::vector<tf::Point>(global_path_.begin() + seg_start_index, global_path_.begin() + seg_end_index);
+    } else {
+        global_segment = std::vector<tf::Point>(global_path_.begin() + seg_start_index, global_path_.end());
+        global_segment.insert(global_segment.end(), global_path_.begin(), global_path_.begin() + seg_end_index);
+    }
+    return global_segment;
+}
+
+double GlobalPath::dtw_distance(const std::vector<tf::Point> &path1, const std::vector<tf::Point> &path2, int w) {
```
Daniel-Martin576, line 97: Maybe put `https://en.wikipedia.org/wiki/Dynamic_time_warping` as a comment to say you got it from there

rr_common/src/planner/global_path.cpp
```diff
@@ -0,0 +1,172 @@
...
+    for (int i = 1; i < n + 1; i++) {
+        for (int j = std::max(1, i - w); j < std::min(m, i + w) + 1; j++) {
+            dtw[i][j] = 0;
+        }
+    }
+
+    for (int i = 1; i < n + 1; i++) {
+        for (int j = std::max(1, i - w); j < std::min(m, i + w) + 1; j++) {
+            double cost = tf::tfDistance(path1[i - 1], path2[j - 1]);
+            dtw[i][j] = cost + std::min(std::min(dtw[i - 1][j], dtw[i][j - 1]), dtw[i - 1][j - 1]);
```
Daniel-Martin576, line 114: Use `std::min({x, y, z});`?

Daniel-Martin576 (changes requested):
Just remove all accepting_updates stuff and some other small stuff.

Daniel-Martin576 (approved):
Looks good! Goob job :+1: I made some small changes that were nitpicks but nothing big

PR #362, opened 2020-11-22
Feat/352-sign-detection

### LINKED ISSUE #352: Sign Detection
**Goal:** Make a node that detects the direction of a sign in an image. Implement this using multi-scale template matching similar to that found [here](https://www.pyimagesearch.com/2015/01/26/multi-scale-template-matching-using-python-opencv/). If you do follow that guide, you will have to convert the code to C++ and make a new thread for each new template. The reason for multi-threading is that template matching can be kinda slow and we want to test multiple templates in parallel. It this case, each template would be a sign pointing in another direction (since template matching is not orientation invariant). Just publish a string:{LEFT, RIGHT, STRAIGHT) to a topic (`/turn_detected`) when a sign is detected. 

**Files Edited (at least):** rr_iarrc/src/sign_detector/sign_detector.cpp

**Success Criteria:**
- Node parameters configurable through the launch file
    - Make the publishing string topic and subscribing image topic configurable 
- Test it on [cloud](https://cloud.robojackets.org/apps/files/?dir=/RoboRacing/Bag%20Files/2019&fileid=122974) urban challenge bag files
- Test with the Urban Challenge Simulation signs. We can add light sources if the signs are too dark.

mittalprakhar:
Closes #352 
- Implements multi-scale template matching (https://www.pyimagesearch.com/2015/01/26/multi-scale-template-matching-using-python-opencv/) with multi-threading
- The scalar parameters (start, end, count) specify the minimum size, maximum size and total number of resizes
- If confidence is above 0.8, a string: {LEFT, RIGHT, STRAIGHT} is published to /turn_detected

See demo [here](https://drive.google.com/drive/folders/10cVtAFw-v6oFcvd_eBcXOdzNex_1PRGZ?usp=sharing).

### DIFF
--- a/rr_iarrc/launch/perception/sign_detector.launch
+++ b/rr_iarrc/launch/perception/sign_detector.launch
@@ -2,34 +2,20 @@
     <node name="sign_detector" pkg="rr_iarrc" type="sign_detector" output="screen">
         <!-- Sign Detector Params -->
         <param name="front_image_subscription" type="string" value="/camera_center/image_color_rect"/>
+        <param name="sign_string_publisher" type="string" value="/turn_detected"/>
 
         <!--Crop Area for detecting signs (-1 keeps whole screen)-->
-        <param name="roi_x" type="int" value="500" />
-        <param name="roi_y" type="int" value="100" />
-        <param name="roi_width" type="int" value="780" />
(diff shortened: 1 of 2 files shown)


Daniel-Martin576 (approved):
Looks good. Good job!

PR #366, opened 2021-01-24
Add 3D Lidar Thresholding

### LINKED ISSUE #357: 3D to 2D lidar scan
**Goal:** Previously, we used a 2D lidar so our mapping stack took in a 2D lidar scan, but now we want to use a 3D lidar. Either add a small node than takes in the 3D lidar scan and thresholds it at a fixed z/height-value or see if our stack can somehow take in a 3D scan to add obstacles (we used costmap_2d). The first option is more likely what is going to end up happening, but you can maybe test the second one. 

**Files Edited (at least):** rr_evgp/src/lidar_z_threshold/lidar_z_threshold.cpp

**Success Criteria:**
- Node parameters configurable through the launch file
- Can make a 2D map like before using the 3D lidar

wbeard01:
-Implemented pcl_ros passthrough nodelet from http://wiki.ros.org/pcl_ros/Tutorials/PassThrough%20filtering.
-Min and max z values configurable through the nodelet's launch file lidar_z_threshold.launch.
-Included the nodelet's launch file in mapping_sim.launch.

Example thresholding can be found [here](https://drive.google.com/drive/folders/121lKrUXqV3nXREwL5xyrjmqNuPSjxSk_?usp=sharing).

Resolves #357.

### DIFF
--- a/rr_evgp/conf/costmap_2d_basic_mapping.yaml
+++ b/rr_evgp/conf/costmap_2d_basic_mapping.yaml
@@ -22,8 +22,8 @@ costmap:
     track_unknown_space: true
     observation_sources: laserscan
     laserscan:
-      data_type: LaserScan
-      topic: /scan
+      data_type: PointCloud2
+      topic: /lidar_thresholded
       frame_name: lidar
       expected_update_rate: 100
       observation_persistence: 0.0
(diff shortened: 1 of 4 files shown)


Daniel-Martin576 (approved):
Looks good. Nice job! :+1:

PR #370, opened 2021-02-01
makes the local map larger

NicoBartholomai:
Makes the local map 3.5 times larger so that it always covers the entire track. This helps prevent the car from trying to drive through a wall.


Daniel-Martin576 (approved):
Looks good!

PR #372, opened 2021-02-02
Removes Urban Bug 

Daniel-Martin576:
I just deleted the actor script for the traffic light for now so that people can use the launch file to test


NicoBartholomai:
So the inline actor script for the startlight was what caused the crash?

Daniel-Martin576:
Seems so. When I removed it, it worked fine.

NicoBartholomai (approved):
Looks good to me, we can find another solution to the stoplight as needed I suppose.

PR #373, opened 2021-02-07
Add URC Turn Action

### LINKED ISSUE #349: Turning Action
**Goal:** When we arrive at a stop bar for the urban challenge, we need to be able to turn left, turn right, and go straight. Make a **_server_** node that takes in a string:{LEFT, RIGHT, STRAIGHT}, does the appropriate action, and returns. This node will be called when we are around the stop bar so you can test it in simulation. You can take inspiration from our [old urc controller](rr_iarrc/src/urc_controller/urc_controller.cpp), but don't rely on it too much. This was previously done by turning at a fixed turning radius until the IMU detected we did a full 90-degree turn. 

**New Files Created (at least):** rr_iarrc/src/urc_turn_action/urc_turn_action.cpp
(If you think of a better name, you can change it)

**Success Criteria:**
-  Node parameters configurable through the launch file
     - Can control turning left_turn_steering, right_turn_steering, turn_speed
     - Make the server name configurable
- Works in Urban Challenge Simulation on multiple intersections

**Future Work** (if you want to give it a try, but don't waste too much time on it):
- We might not be approaching the stop bar head-on (maybe at an angle). Adapt your node so if the stop bar gives you the car's relative position to the stop bar that you can still successfully make a turn.

tobynguyen03:
- Created/included new turning.srv file and urc_turn_action subdirectory
- Turn direction and approach angle configurable through /turn_action service
- Steering parameters adjusted to account for wider left turn and sharper right turn
- Demo video: https://drive.google.com/drive/u/1/folders/1cbXxb3keX6QeDnX0wQRj6z2AtA1LejRY

Resolves #349


Daniel-Martin576 (approved):
I just formated it and deleted the turning_test folder for you. Looks good. Good Job! :+1:

PR #376, opened 2021-02-14
Optimize over speed and steering 

Daniel-Martin576:
Changes the planner to optimize speed and steering. Modified DIstance Map so that it can be more flexible. Changed local map to use lidar data directly. 

[Demo Video](https://drive.google.com/file/d/1TJFcsj6eEeVwT6XMZn2jjWcU5fT4oCvO/view?usp=sharing)

I updated:
- Distance Map (remove if-updating stuff, and add nonlethal layer logic)
- Bicycle model
- Rviz (not sure if now is better or worse)
- Added reverse speed variable for recovery
- Added path width variable 
- Some plan config parameters


Daniel-Martin576:
Has some issues with opponent cars, but we still need to work on that

rr_evgp/conf/planner_sim.yaml
```diff
@@ -15,10 +15,10 @@ steering_filter:
 
 speed_filter:
     init_val: 0
-    val_max: 13.5
-    val_min: -1.0
-    rate_max: 20.0
-    rate_min: -25.0
+    val_max: 25
```
NicoBartholomai, line 18: Are you sure this is safe?

NicoBartholomai (approved):
It all checks out for me.

PR #379, opened 2021-02-14
Opponent detection

### LINKED ISSUE #355: Opponent Detection
**Goal:** We need to detect other opponents using the 3D lidar. You can do this by doing wall and ground segmentation then clustering the leftover points or some other ML algorithm. I will let this be a bit more open-ended since there are a few ways to approach the problem. Ideally you should be able to recognize cars across multiples scans (ex. car 4) so we can predict their movements, but let's address car detection first. 

**New Files Created (at least):** rr_evgp/src/opponent_detection/opponent_detection.cpp

**Success Criteria:**
- Can reliably find opponents in front and to the sides of our vehicle
- Test in the evgp gazebo world with multiple opponents (look at `feat/320-adding-opponent-cars` current in PR for review)

jshao109:
Fixes #355 
Finished opponent detection node


Daniel-Martin576 (approved):
Looks good! Good Job :+1: I made some changes just to make the code more readable and added centroids

PR #382, opened 2021-03-02
EVGP World with Cones

jshao109:
Resolves issue #381. Made taller cone model, replaced walls of EVGP track with cones


Daniel-Martin576 (changes requested):
For `rr_gazebo/worlds/ev_grand_prix_cones.world` please substitute the cone models with `<include>`'s similar to the other world files and remove the car model for being defined in the world. Other than that looks good!

Daniel-Martin576 (approved):
Looks good! Good job :+1:  I uncommented some stuff in the launch file to actually get the car to spawn.

PR #389, opened 2021-04-11
Feat/387 path progress

### LINKED ISSUE #387: Path Progress
Add a progress metric to encourage the car to drive forward as much as it can.

NicoBartholomai:
Closes #387. Adds a path progression heuristic to the planner. For each projected path created, it finds the points on the global path that are closest to the beginning and ending of that projected path, and then finds the distance of this global path segment. This encourages longer paths to be generated, and thus encourages the bot to move forward.

This can be tested by roslaunching rr_gazebo evgp.launch and rr_evgp evgp_planning_demo.launch
Visualizing the /progress_path_seg topic will show the current global path segment the plan is using.


NicoBartholomai:
Re do for ROS 2

PR #393, opened 2021-08-22
Feat/lidar car

idale0001:
Updated opponent detection files so that the opponents are more consistently assigned the same id and color by using a greedy algorithm to find the clusters that are closest to the previous clusters.

### DIFF
--- a/rr_evgp/src/opponent_detection/opponent_detection.cpp
+++ b/rr_evgp/src/opponent_detection/opponent_detection.cpp
@@ -59,23 +59,54 @@ double euclidDistance(geometry_msgs::Point p1, geometry_msgs::Point p2) {
     return sqrt(x2 + y2 + z2);
 }
 void trackMovement(const std::vector<geometry_msgs::Pose>& centroids, visualization_msgs::MarkerArray& marker_array) {
-    int indexOfPrev = 0;
-    for (geometry_msgs::Pose prevCentroid : prevCentroids) {
-        double shortestDistance = euclidDistance(prevCentroid.position, centroids[0].position);
-        int indexOfClosest = 0;
-        for (int i = 1; i < centroids.size(); i++) {
-            double distance = euclidDistance(prevCentroid.position, centroids[i].position);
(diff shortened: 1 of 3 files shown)


NicoBartholomai (changes requested):
There is a faster way to find pairs between two sets but I think since they're so small this is fine. You could consider creating your own datatype that just holds a pointer to the clusters and their respective centroids instead of trying to keep track of the same index within lists.

rr_evgp/src/opponent_detection/opponent_detection.cpp
```diff
@@ -47,8 +52,78 @@ visualization_msgs::Marker makeMarkers(const std::vector<geometry_msgs::Point>&
...
+    int largest_id = 0;
+    std::unordered_set<int> ids;
+    for (int i = 0; i < prevCentroids.size(); i++) {
+        std::vector<double> row;
+        if (ids.find(prevMarkers.markers[i].id) == ids.end()) {
+            ids.insert(prevMarkers.markers[i].id);
+        }
+        if (prevMarkers.markers[i].id > largest_id) {
+            largest_id = prevMarkers.markers[i].id;
+        }
```
Daniel-Martin576, line 75: max

rr_evgp/src/opponent_detection/opponent_detection.cpp
```diff
@@ -47,8 +52,78 @@ visualization_msgs::Marker makeMarkers(const std::vector<geometry_msgs::Point>&
...
+        for (geometry_msgs::Pose centroid : centroids) {
+            double distance = euclidDistance(prevCentroids[i].position, centroid.position);
+            row.push_back(distance);
+        }
```
Daniel-Martin576, line 79: Could do transform if wanted to be fancy, but not nessessary

Daniel-Martin576: Const reference

NicoBartholomai:
This will be re-implemented in ROS 2

PR #398, opened 2021-11-07
ROS2 rr_description

NicoBartholomai:
Refactors rr_desctiption to run in ros 2. 
Adds the  ROS2 branch of the velodyne simulator as a third party package so that the simulated velodyne is properly loaded.


idale0001 (approved):
LGTM

PR #428, opened 2021-11-22
Ros2/create documentation

thecharlesjenkins:
First draft of documentation. I will continue to work on this but I want to make sure that new members can start to look at it as they develop their first issues. I also created the Github Actions to get it to build automatically and figured out how the `if` statement works. Here is the url: https://robojackets.github.io/roboracing-software/.

I also created packages for all of the packages we talked about creating.


docs/source/installation.rst
```diff
@@ -0,0 +1,147 @@
...
+    
+    sudo apt-get install doxygen
+    pip install breathe
+
+Step 4 - Build Documentation
+----------------------------
+
+.. code-block::
+
+    cd docs
```
idale0001, line 140: might want to change cd docs to an absolute path so people can just copy and paste the commands easier. This seems to assume that they are in the roboracing directory but they might not be if they are only looking at how to build the documentation and have already downloaded the repo. Its also possible they had to change directories because of issues with other steps

docs/source/installation.rst
```diff
@@ -0,0 +1,147 @@
...
+
+Download the file into ``~/java/`` and rename the file to plantuml.jar. The download path must be: ``~/java/plantuml.jar``
+
+Optional:
+
+To enable plantuml from anywhere in your file structure add the following line to your bashrc. :doc:`linux_info/bashrc`
+
+.. code-block::
+
+    echo alias plantuml="'java -jar /home/<username>/java/plantuml.jar'" >> ~/.bashrc
```
idale0001, line 116: Why does it say to use /home/<username> instead of ~/? Not necessary to change I'm just curious. Though maybe we should change ~/java with /home/<username> because some people might not know what ~ means

docs/source/resources/committing.rst
```diff
@@ -0,0 +1,15 @@
...
+============================
+
+1. Make sure you followed the :doc:`Installation Guide </source/installation>`
+2. Go to ClickUp.com
+    * If you are not already signed up, navigate to my.robojackets.com, go to the profile and add an email to the ClickUp field.
+3. Find the issues assigned to you
+4. Move the task to "In Progress" when you start working on it
+5. Create a branch name of the type: **feat/short_issue_title**
+    * Give your issue some good name that is not too long
+6. Once you finish your issue, create a pull request: https://github.com/RoboJackets/roboracing-software/pulls
```
idale0001, line 12: Should we require a link to the clickup issue as part of the pull request description? Probably would make it easier to review it does everything its suppose to

rr_demo/package.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+  <name>rr_demo</name>
+  <version>0.0.0</version>
+  <description>TODO: Package description</description>
```
idale0001, line 6: You can probably fill in the TODOs now

rr_msgs/package.xml
```diff
@@ -3,6 +3,14 @@
 <package format="3">
   <name>rr_msgs</name>
   <version>0.0.0</version>
+<<<<<<< HEAD
+  <description>TODO: Package description</description>
```
idale0001, line 7: Think this is also in a few other places, but it looks like there are some places where GitHub is saying that both versions of merge conflicts are going to be added so make sure that the proper version Is being merged into main

idale0001 (changes requested):
Mostly looks good just a few small things. Also it looks like it is not currently building according to github actions so if you could figure out why that is that would be good too.

.github/workflows/workflow.yml
```diff
@@ -22,4 +22,17 @@ jobs:
         inplace: True
     - uses: EndBug/add-and-commit@v4
       with:
-        message: 'fix format'
\ No newline at end of file
+        message: 'fix format'
+
+  build_documentation:
```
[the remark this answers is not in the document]
thecharlesjenkins: I will remove it from this file and only include it in the other.

.gitignore
```diff
@@ -1,3 +1,6 @@
+.vscode
```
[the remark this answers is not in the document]
thecharlesjenkins: Oops I will remove these two lines

docs/source/installation.rst
thecharlesjenkins: I will write a note explaining what ~ is

docs/source/installation.rst
thecharlesjenkins: Yeah that's a good idea

docs/source/resources/committing.rst
thecharlesjenkins: Yes I like that

rr_demo/package.xml
thecharlesjenkins: Sounds good

rr_msgs/package.xml
thecharlesjenkins: Oh oops some merge conflicts that I missed

idale0001 (approved):
Looks great!

PR #439, opened 2022-01-25
Ros2/create util

idale0001:
Start adding files to rr_util, specifically a node for the joystick so we can manually control the robot even if electrical has issues.


thecharlesjenkins (changes requested):
It looks great! One thing that might be nice is to add a link to the launch file and the command to run it `ros2 launch rr_util joystick_driver.launch.xml`

thecharlesjenkins (approved):
Looks good!

